PR #6, opened 2023-05-24
Multiple languages

helixbass:
In this PR:
- add support for targeting Rust vs Typescript source files/queries

To test:
Existing behavior should still work except you should have to add an explicit `-l rust` or `--language rust` command-line argument
If you add eg a file named `tmp.ts` with contents `interface Foo {}` in the project root directory and then run `cargo run -- -l typescript -q "(interface_declaration) @interface_declaration"` you should see a result for that interface declaration in that source file

### DIFF
--- a/src/language.rs
+++ b/src/language.rs
@@ -1,0 +1,54 @@
+use clap::ValueEnum;
+use tree_sitter::Language;
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, ValueEnum)]
+pub enum SupportedLanguageName {
+    Rust,
+    Typescript,
+}
+
+impl SupportedLanguageName {
+    pub fn get_language(self) -> Box<dyn SupportedLanguage> {
+        match self {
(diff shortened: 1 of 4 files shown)


src/language.rs
```diff
@@ -0,0 +1,54 @@
...
+
+pub fn get_rust_language() -> SupportedLanguageRust {
+    SupportedLanguageRust
+}
+
+pub struct SupportedLanguageTypescript;
+
+impl SupportedLanguage for SupportedLanguageTypescript {
+    fn language(&self) -> Language {
+        tree_sitter_typescript::language_tsx()
```
helixbass, line 44: Looks like tsx vs typescript are two different tree-sitter grammars, ultimately I'd guess the nicest ergonomics would be to eg try and parse the given query as both (and invoke the appropriate per-grammar parsed query based on each source file's extension?) and if it can't parse it for whichever grammar then silently skip those files?

PR #7, opened 2023-05-28
Use grep crates

peterstuart:
This is an initial (extremely hacky attempt) at using ripgrep's full infrastructure via it's `grep` crate (which wraps a few other crates). It has a bug right now where it always prints out matches as the first lines of the file the match is in, eg:

```
cargo run -- --language rust --query-source "(field_declaration_list (field_declaration type: (type_identifier) @type))"
   Compiling tree-sitter-grep v0.1.0 (/home/peter/dev/personal/tree-sitter-grep)
    Finished dev [unoptimized + debuginfo] target(s) in 1.61s
     Running `target/debug/tree-sitter-grep --language rust --query-source '(field_declaration_list (field_declaration type: (type_identifier) @type))'`
./src/lib.rs:1:use clap::Parser;
./src/lib.rs:2:use grep::matcher::Match;
```

Assuming that's fixable, using all these crates seems potentially pretty great, even despite the weird mismatch between the `grep` crates expected regex-style "matchers".

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,8 +1,15 @@
 use clap::Parser;
+use grep::matcher::Match;
+use grep::matcher::Matcher;
+use grep::matcher::NoCaptures;
+use grep::matcher::NoError;
+use grep::searcher::Searcher;
 use ignore::{types::TypesBuilder, DirEntry, WalkBuilder};
 use rayon::prelude::*;
 use std::fs;
+use std::io;
 use std::path::PathBuf;
+use std::sync::Mutex;
 
 mod language;
 mod macros;
@@ -42,19 +49,20 @@ pub fn run(args: Args) {
             .capture_index_for_name(capture_name)
             .expect(&format!("Unknown capture name: `{}`", capture_name))
     });
+
     enumerate_project_files(&*supported_language)
         .par_iter()
-        .flat_map(|project_file_dir_entry| {
-            get_results(
-                &query,
-                project_file_dir_entry.path(),
-                capture_index,
-                language,
-            )
+        .for_each(|project_file_dir_entry| {
+            let mut printer = grep::printer::Standard::new_no_color(io::stdout());
+            let path = project_file_dir_entry.path();
+            let matches = get_results(&query, path, capture_index, language);
+
+            let matcher = TreeSitterMatcher::new(matches);
+
+            Searcher::new()
+                .search_path(&matcher, path, printer.sink_with_path(&matcher, path))
+                .unwrap();
         })
-        .for_each(|result| {
-            println!("{}", result.format());
-        });
 }
 
 fn enumerate_project_files(language: &dyn SupportedLanguage) -> Vec<DirEntry> {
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -42,19 +49,20 @@ pub fn run(args: Args) {
...
         .par_iter()
-        .flat_map(|project_file_dir_entry| {
-            get_results(
-                &query,
-                project_file_dir_entry.path(),
-                capture_index,
-                language,
-            )
+        .for_each(|project_file_dir_entry| {
+            let mut printer = grep::printer::Standard::new_no_color(io::stdout());
```
peterstuart, line 56: Prints out the matches, by default in a "normal" grep style, I guess, but is configurable via a builder to handle all the standard grep output options.

src/lib.rs
```diff
@@ -42,19 +49,20 @@ pub fn run(args: Args) {
...
+            let path = project_file_dir_entry.path();
+            let matches = get_results(&query, path, capture_index, language);
+
+            let matcher = TreeSitterMatcher::new(matches);
```
peterstuart, line 60: We need something to conform to the `Matcher` trait. This seems stupid because it's a trait about incrementally searching through the bytes of a string, but this is really just faking it by getting all results from the file and then feeding them out to the searching via the `Matcher` interface.

src/lib.rs
```diff
@@ -42,19 +49,20 @@ pub fn run(args: Args) {
...
+
+            Searcher::new()
+                .search_path(&matcher, path, printer.sink_with_path(&matcher, path))
+                .unwrap();
```
peterstuart, line 64: I don't think this is really getting us much of anything, since it's really meant to deal with incremental searching in a smart way, but it seemed like the easiest way to use the printer, given the printer's interface.

src/lib.rs
```diff
@@ -72,3 +80,37 @@ fn enumerate_project_files(language: &dyn SupportedLanguage) -> Vec<DirEntry> {
...
+    type Error = NoError;
+
+    fn find_at(&self, _haystack: &[u8], at: usize) -> Result<Option<Match>, Self::Error> {
+        let _match = self
+            .matches
+            .lock()
+            .unwrap()
+            .pop()
+            .map(|m| Match::new(m.start() - at, m.end() - at));
+        Ok(_match)
```
peterstuart, line 110: The function params are:
- `haystack`: remaining bytes to search
- `at`: offset into the file where `haystack` starts, I think

It wants matches returned relative to the start of the haystack, so subtrack `at` from the "real" matches. (`at` has always been 0 in my experiments, though)

src/lib.rs
```diff
@@ -72,3 +80,37 @@ fn enumerate_project_files(language: &dyn SupportedLanguage) -> Vec<DirEntry> {
...
+    }
+
+    fn new_captures(&self) -> Result<Self::Captures, Self::Error> {
+        Ok(NoCaptures::new())
```
peterstuart, line 114: I don't even know what captures do in regular grep.

src/treesitter/mod.rs
```diff
@@ -58,19 +33,9 @@ pub fn get_results(
...
-                        .lines()
-                        .nth(point.row)
-                        .unwrap()
-                        .unwrap()
-                },
-                file_path: file_path.to_owned(),
-            }
+            let range = node.range();
+
+            Match::new(range.start_byte, range.end_byte)
```
peterstuart, line 38: Directly use the `Match` struct, even though the start and end numbers are potentially wrong (see my comment above about `at`).

src/lib.rs
helixbass: I think captures are when you have `(...)` in your regex pattern and then can reference those specific chunks of the matched text in eg a `.replace()` "replacer" argument

helixbass (approved):
@peterstuart nice, I opened https://github.com/peterstuart/tree-sitter-grep/pull/1 against this PR/branch, feel free to merge it if/when you think it's ready (and then I sent you some sort of access invite for this repo, that should at least be a step in the right direction for giving you "write" permission so that you can eg merge this PR)

src/lib.rs
peterstuart: Oh yeah, I knew about regex captures in general, but I didn't know grep could do things with them (can it do string replacements?)

helixbass: Ah right I'm conflating `regex`/`ripgrep`, I'm not sure what capture-related "surface area" `ripgrep` exposes?

PR #8, opened 2023-05-29
Load dynamic library for matching

helixbass:
In this PR:
- change to a Cargo "workspace" so that a "shared" crate for "plugin-y stuff" can be separated out from the main crate
- expose a "Rust-to-Rust FFI" interface for specifying a path to a shared library that exports a "plugin" whose job is to filter potential matches (at the moment an initial tree-sitter query is still required but should probably be optional if you supply a dynamic-library filter (and the filter would just get passed every single tree-sitter AST node)?) (and optionally a string command-line argument to "forward" to the loaded plugin to "configure" it)

To test:
At the moment I included an example plugin in the committed code (though I'd plan on removing it before merging) so you should be able to run `cargo build --release --all` from the (workspace) root and then run eg `./target/release/tree-sitter-grep -l rust --filter ./target/release/libfilter_after_line_number.dylib --filter-arg 100 ./reference-parameters.scm` (the path to the built dynamic library may be different (`*.so`?) if you aren't on a Mac?) and see a list of reference-type'd function parameters that occur on line numbers after 100

If you (then) run `./target/release/tree-sitter-grep -l rust --filter ./target/release/libfilter_after_line_number.dylib ./reference-parameters.scm` you should see an ugly error that it expected you to supply a configuring argument to the filter (via the `--filter-arg`/`-a` command-line argument)

If you run `./target/release/tree-sitter-grep -l rust --filter ./target/release/libfilter_after_line_number.dylib -a whee ./reference-parameters.scm` you should see an ugly error that the passed filter-argument argument wasn't of the expected parse-able type (ie `usize`)

You should still be able to invoke without passing a `--filter`/`-f` argument

If you're curious about performance, you should be able to observe (eg via the `time` command?)/it appeared that there was no significant difference in total time (I measured against running queries against the Typescript compiler in Rust codebase) when running with a dynamically-loaded filter vs without

You should be able to create, build + use a different filter (eg one that doesn't accept/require a configuring argument)


tree_sitter_grep/src/plugin.rs
```diff
@@ -0,0 +1,80 @@
...
+    fn register_filterer(&mut self, filterer: Box<dyn Filterer>) {
+        let proxy = FiltererProxy {
+            filterer,
+            _lib: self.lib.clone(),
+        };
+        self.filterer = Some(proxy);
+    }
+}
+
+fn load_plugin(library_path: impl AsRef<OsStr>, filter_arg: Option<&str>) -> FiltererProxy {
```
helixbass, line 45: I think this should only be getting called once (across the multiple threads running the query + then filtering) if I understand `once_cell` correctly (ie all of those threads should be able to "compete" for who gets to initialize the `OnceCell` but only one of them "wins" and then the others' `.get_or_init()` calls would block until the one is done initializing)? But I didn't sanity-check that

tree_sitter_grep/src/plugin.rs
```diff
@@ -0,0 +1,80 @@
...
+    let library = Arc::new(unsafe {
+        Library::new(library_path).expect("Couldn't load filter dynamic library")
+    });
+
+    let plugin_declaration = unsafe {
+        library
+            .get::<*mut PluginDeclaration>(b"plugin_declaration\0")
```
helixbass, line 52: Ok so this is again mimicking that tutorial. Other things I was reading were [this](https://nullderef.com/blog/plugin-start/) and the corresponding [Reddit discussion](https://www.reddit.com/r/rust/comments/q2n6b8/plugins_in_rust_diving_into_dynamic_loading/) - they warn that "Rust-to-Rust FFI" really technically isn't safe at all (even with sanity-checking that you're using the same version of the Rust compiler etc at dynamic-library-loading time) because the Rust compiler reserves the right to eg change struct layouts _between compiler runs_, but that it sounds like (at least currently) it never really does that in practice

That series looks helpful and well-written in general, it's doing some exploring/explaining of alternatives eg WASM

tree_sitter_grep_plugin_core/build.rs
```diff
@@ -0,0 +1,4 @@
+fn main() {
+    let version = rustc_version::version().unwrap();
+    println!("cargo:rustc-env=RUSTC_VERSION={}", version);
```
helixbass, line 3: Per that tutorial, I don't really understand what "build scripts" are/do?

tree_sitter_grep_plugin_core/src/lib.rs
```diff
@@ -0,0 +1,33 @@
+use tree_sitter::Node;
+
+// per https://adventures.michaelfbryan.com/posts/plugins-in-rust/
+
+pub trait Filterer: Send + Sync {
```
helixbass, line 5: It seemed like I needed to add this `Send + Sync` bound in order to get the (`static`) `OnceCell` to be happy?

PR #9, opened 2023-05-30
Filter dynamic library: simple C FFI

helixbass:
In this PR:
- as discussed, update the FFI interface to be C-compatible (and simpler)

To test:
I included an example filter in this branch again (temporarily) so you should be able to run `cargo build --release --all` and then run eg `./target/release/tree-sitter-grep -l rust --filter ./target/release/libtree_sitter_grep_filter_before_line_number.dylib --filter-arg 30 ./reference-parameters.scm` and see matches printed that only have line numbers < 30
If you run that without `--filter-arg` passed then it should fail on expecting to be passed a filter argument
If you run that with eg `--filter-arg abc` passed then it should fail on expecting to be passed a usize filter argument
If in the example filter you comment out the `initialize()` function (and the static atomic) and just put a hardcoded line number inside the `filterer()` callback eg `20` and re-build and re-run the above (without necessarily including the `--filter-arg 30`) then you should see matches printed that only have line numbers < 20
Running without a `--filter` argument should still work as expected
If you're feeling bold, try writing a filter plugin in a different language and seeing if it works?

Based on `match-predicate` (Maybe looking at the diff of this branch vs current `master` would be preferable since this branch tears out a lot of what was on `match-predicate`, feel free to change the base branch of this PR)


src/plugin.rs
```diff
@@ -0,0 +1,69 @@
...
+        unsafe { (self.filterer)(node) }
+    }
+}
+
+fn load_plugin(library_path: impl AsRef<OsStr>, filter_arg: Option<&str>) -> Filterer {
+    let library =
+        unsafe { Library::new(library_path).expect("Couldn't load filter dynamic library") };
+
+    if let Some(initialize) =
+        unsafe { library.get::<unsafe extern "C" fn(*const libc::c_char)>(b"initialize\0") }.ok()
```
helixbass, line 29: Ok so the plugin "API" I went with is:
Optional `initialize()` function that takes a C-style string (so plugins that don't need to be "configured" by the `--filter-arg` command-line argument can omit `initialize()` entirely)
Specifically if no `--filter-arg` argument is passed but `initialize()` is implemented then it gets passed a null pointer

`filterer()` function that takes what should effectively be a pointer to a C tree-sitter `Node` and returns a bool

src/plugin.rs
```diff
@@ -0,0 +1,69 @@
...
+                initialize(ptr::null());
+            }
+        }
+    }
+
+    let filterer = unsafe {
+        library
+            .get::<unsafe extern "C" fn(*const Node) -> bool>(b"filterer\0")
+            .expect("Couldn't load expected symbol from filter dynamic library")
+            .into_raw()
```
helixbass, line 52: Per above comment this converts from the "safe non-platform-specific" `libloading::Symbol` type to the "unsafe (ie not dependent on the lifetime of the parent `Library`) platform-specific" type that I aliased as `RawSymbol`

src/plugin.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+}
+
+pub fn get_loaded_filter(
+    filter_library_path: Option<&str>,
+    filter_arg: Option<&str>,
+) -> Option<&'static Filterer> {
+    filter_library_path.map(|filter_library_path| {
+        static LOADED_FILTERER: OnceCell<Filterer> = OnceCell::new();
+        LOADED_FILTERER.get_or_init(|| load_plugin(filter_library_path, filter_arg))
```
helixbass, line 67: Again haven't sanity-checked that this initialization callback only happens once (across all threads)

tree_sitter_grep_filter_before_line_number/src/lib.rs
```diff
@@ -0,0 +1,25 @@
+use std::{
+    ffi::CStr,
+    sync::atomic::{AtomicUsize, Ordering},
+};
+
+use libc::c_char;
+use tree_sitter::Node;
+
+static ROW_NUMBER: AtomicUsize = AtomicUsize::new(0);
```
helixbass, line 9: Most of this stuff is only necessary to wire up the configuration argument

Eg here I went with a `static` `AtomicUsize` for (globally, which the API seems to force you to do, which seems fine?) remembering the passed configuration argument

tree_sitter_grep_filter_before_line_number/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[no_mangle]
+pub extern "C" fn initialize(value: *const c_char) {
+    assert!(!value.is_null(), "Expected filter argument");
+    let value: usize = unsafe { CStr::from_ptr(value) }
```
helixbass, line 14: This is per eg [this](http://jakegoulding.com/rust-ffi-omnibus/string_arguments/) (for how to accept a string argument via FFI)

So the "ownership contract" here would seem to be that the plugin can't assume that this passed C string outlives the `initialize()` call (ie it doesn't own the passed string)? And here we shouldn't be violating that since we convert to an (owned) `usize`

src/plugin.rs
```diff
@@ -0,0 +1,69 @@
+use libloading::Library;
+use once_cell::sync::OnceCell;
+use std::ffi::{CString, OsStr};
+use std::ptr;
+use tree_sitter::Node;
+
+#[cfg(unix)]
+type RawSymbol<TValue> = libloading::os::unix::Symbol<TValue>;
+#[cfg(windows)]
+type RawSymbol<TValue> = libloading::os::windows::Symbol<TValue>;
```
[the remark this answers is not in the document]
peterstuart: I'd find this slightly more clear if `TValue` was `TCFunction` or something like that.

tree_sitter_grep_filter_before_line_number/src/lib.rs
peterstuart: Ah oh, that makes sense

src/plugin.rs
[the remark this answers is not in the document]
helixbass: I think technically `TCFunction` is too "narrow" because exported symbols can also be things like globals, but yes `TValue` is a terrible name so I updated to `TExportedSymbol`

src/plugin.rs
helixbass: Woof ok I know that the expected type is `&[u8]` so I guess `b"..."` ("byte strings"?) gives you an `&[u8]` rather than a `&str`?

The `\0` I saw in some examples (but not others) and then saw in the `libloading` docs that "Providing a null-terminated `symbol` may help to avoid an allocation."

Ok yes looks like it's basically [ASCII-encoding](https://doc.rust-lang.org/reference/tokens.html#byte-string-literals)

tree_sitter_grep_filter_before_line_number/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+        .unwrap()
+        .parse()
+        .expect("Expected filter argument to be a usize");
+    ROW_NUMBER.store(value, Ordering::Relaxed);
+}
+
+#[no_mangle]
+pub extern "C" fn filterer(node: &Node) -> bool {
+    node.start_position().row < ROW_NUMBER.load(Ordering::Relaxed)
+}
```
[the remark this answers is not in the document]
helixbass: I like that idea and does seem worth trying to make plugin-writing as easy as possible

peterstuart (approved):
Ok, tried this out for real and it's working for me.

Cargo.toml
```diff
@@ -1,9 +1,28 @@
...
+tree-sitter-typescript = "0.20.2"
+
+[lib]
+name = "tree_sitter_grep"
+
+[[bin]]
+name = "tree-sitter-grep"
+
+[workspace]
+members = ["tree_sitter_grep_filter_before_line_number"]
```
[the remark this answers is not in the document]
helixbass: Removed the example filter before merging, pushed it as a [separate repo](https://github.com/helixbass/tree_sitter_grep_filter_before_line_number)

src/plugin.rs
```diff
@@ -0,0 +1,63 @@
...
+        unsafe { library.get::<unsafe extern "C" fn(*const libc::c_char)>(b"initialize\0") }.ok()
+    {
+        let filter_arg = filter_arg.map(|filter_arg| {
+            CString::new(filter_arg).expect("Couldn't convert provided filter arg to CString")
+        });
+        unsafe {
+            initialize(
+                filter_arg
+                    .as_ref()
+                    .map_or_else(|| ptr::null(), |filter_arg| filter_arg.as_ptr()),
```
helixbass, line 37: > I don't understand why this isn't equivalent to the below but when I tried this it was passing an empty (C) string to the plugin?

Ok as discussed the issue with how I had this was presumably that the `.map_or_else()` was consuming `filter_arg` and thus the `.as_ptr()` was returning a dangling pointer. So in theory then doing `filter_arg.as_ref().map_or_else(...)` should work (since `filter_arg` is still alive, so the pointer into it should still be valid), and I tested it out and does appear to work correctly

PR #10, opened 2023-05-30
Parallelize enumerate project files -> query matching

helixbass:
In this PR:
- instead of "blocking" on `.collect()`'ing all enumerated project files (via `ignore`) into a `Vec` and _then_ proceeding into the (parallelized) query-matching phase, parallelize the enumerating of project files and start "flowing" those through into query-matching as they are enumerated

I did this because it was seeming like there was a noticeable initial "pause" before any results started showing up and I assumed that was the "enumerating project files" phase, which made me think that it would probably be a significant improvement to overall runtime to not block on that. Indeed here is a basic timing graph from before this branch:
Screen Shot 2023-05-30 at 12 21 48 PM
And after:
Screen Shot 2023-05-30 at 12 22 13 PM

In terms of roughly-measured (via `time`) overall runtime, against the Typescript compiler in Rust codebase it looks like it went from ~460-480ms to ~280-300ms so maybe a 35-40% speedup?

Based on `filter-plugin-c-ffi`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,5 +28,5 @@ name = "tree-sitter-grep"
 [workspace]
 members = ["tree_sitter_grep_filter_before_line_number"]
 
-[profile.release]
-debug = true
+# [profile.release]
+# debug = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,9 +11,10 @@ grep = "0.2.12"
 ignore = "0.4.20"
 libc = "0.2.144"
 libloading = "0.8.0"
+num_cpus = "1.15.0"
 once_cell = "1.17.1"
-rayon = "1.7.0"
 regex = "1.8.2"
+threadpool = "1.8.1"
 tree-sitter = "0.20.10"
 tree-sitter-rust = "0.20.3"
 tree-sitter-typescript = "0.20.2"
@@ -26,3 +27,6 @@ name = "tree-sitter-grep"
 
 [workspace]
 members = ["tree_sitter_grep_filter_before_line_number"]
+
+[profile.release]
+debug = true
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,12 +4,17 @@ use grep::matcher::Matcher;
 use grep::matcher::NoCaptures;
 use grep::matcher::NoError;
 use grep::searcher::SearcherBuilder;
+use ignore::WalkState;
 use ignore::{types::TypesBuilder, DirEntry, WalkBuilder};
-use rayon::prelude::*;
 use std::cell::RefCell;
 use std::fs;
 use std::io;
 use std::path::PathBuf;
+use std::sync::mpsc;
+use std::sync::mpsc::Sender;
+use std::sync::Arc;
+use std::thread;
+use threadpool::ThreadPool;
 use tree_sitter::{Language, Query};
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -26,3 +27,6 @@ name = "tree-sitter-grep"
 
 [workspace]
 members = ["tree_sitter_grep_filter_before_line_number"]
+
+# [profile.release]
```
helixbass, line 31: I just had this set per the [`cargo-instruments`](https://crates.io/crates/cargo-instruments) docs, could remove

src/lib.rs
```diff
@@ -50,36 +55,48 @@ pub fn run(args: Args) {
...
-                args.filter_arg.clone(),
-            );
-
-            SearcherBuilder::new()
-                .multi_line(true)
-                .build()
-                .search_path(&matcher, path, printer.sink_with_path(&matcher, path))
-                .unwrap();
-        })
+    let (sender, receiver) = mpsc::channel::<DirEntry>();
```
helixbass, line 65: Ok so the basic "thread architecture" here is:
Spawn a "receiver" thread whose job is to get project file `DirEntry`'s from a channel and for each one add a job to a thread pool (for doing the query-matching for that file)

And in the main thread do a parallelized enumerating of the project files and send each one on the channel as they get enumerated

src/lib.rs
```diff
@@ -50,36 +55,48 @@ pub fn run(args: Args) {
...
+
+                    SearcherBuilder::new()
+                        .multi_line(true)
+                        .build()
+                        .search_path(&matcher, path, printer.sink_with_path(&matcher, path))
+                        .unwrap();
+                }
+            });
+        }
+        thread_pool.join();
```
helixbass, line 93: This blocks on all of the thread pool jobs finishing so therefore the "receiver" thread shouldn't return until (a) the project file enumeration has completed (at which point the sender should get dropped and `receiver.recv()` should then return an `Err`) and (b) all of the thread-pool query-matching jobs have run to completion

src/lib.rs
```diff
@@ -50,36 +55,48 @@ pub fn run(args: Args) {
...
+    });
+    walk_project_files(&*supported_language, sender);
+    handle.join().unwrap();
```
helixbass, line 96: ...so `.join()`'ing that "receiver" thread should suffice for not "killing the program (by falling off the end of `main()`) before everything has actually finished"?

src/lib.rs
```diff
@@ -88,11 +105,23 @@ fn enumerate_project_files(language: &dyn SupportedLanguage) -> Vec<DirEntry> {
                 .build()
                 .unwrap(),
         )
-        .build()
-        .into_iter()
-        .filter_map(|entry| entry.ok())
-        .filter(|entry| entry.metadata().unwrap().is_file())
-        .collect()
+        .build_parallel()
```
helixbass, line 108: This gives you a [`WalkParallel`](https://docs.rs/ignore/latest/ignore/struct.WalkParallel.html) (instead of a `Walk`) which rather than giving you an iterator-like interface has you supply a callback to `.run()` that (returns a callback that) gets invoked for each enumerated project file

helixbass:
Closing in favor of #11

PR #11, opened 2023-06-06
Use rayon + WalkParallel

helixbass:
In this PR:
- "hand off" from `ignore` `WalkParallel` (for enumerating project files in parallel) to `rayon` (for running the tree-sitter querying in parallel) (via its `.par_bridge()` helper)

To test:
Behavior should still work and performance still appears pretty comparable to the manual channel + thread-pool version in `dont-block-on-enumerate-project-files`

Based on `dont-block-on-enumerate-project-files` (again maybe it would be preferable to see the diff against the "base branch" `filter-plugin-c-ffi`, feel free to change it)


src/lib.rs
```diff
@@ -62,41 +65,88 @@ pub fn run(args: Args) {
...
+                SearcherBuilder::new()
+                    .multi_line(true)
+                    .build()
+                    .search_path(&matcher, path, printer.sink_with_path(&matcher, path))
+                    .unwrap();
+            }
+        });
+}
+
+trait IntoParallelIterator {
```
helixbass, line 92: Expressing this via a trait was wholly unnecessary (I [initially](https://github.com/helixbass/tree-sitter-grep/pull/11/commits/a5c4cbc7bfdb1f85046ad848b882135a99d3d0e7) had it as
```
WalkParallelIterator::new(get_project_file_walker(&*supported_language))
    .par_bridge()
    .for_each(...)
```
), I just thought the `.into_parallel_iterator().for_each(...)` chaining (+ encapsulating of `.par_bridge()`) was stylish

src/lib.rs
```diff
@@ -62,41 +65,88 @@ pub fn run(args: Args) {
...
+}
+
+impl Iterator for WalkParallelIterator {
+    type Item = DirEntry;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.receiver_iterator.next() {
+            Some(item) => Some(item),
+            None => {
+                // self.handle.join().unwrap();
```
helixbass, line 142: Not sure what my mental model is/should be here - we've definitely "exhausted" this encapsulated thread because the owned receiver iterator is returning `None` so therefore all senders have been dropped (ie the `walk_parallel.run()` callback has been dropped) so therefore `walk_parallel` must basically be done and most likely has already returned (ie the thread has already finished)? So then this `.join()` shouldn't block for long if at all? But by the same token is there any reason to explicitly `.join()` at that point (my understanding is that if the main thread finished `main()` before that thread had actually finished then it would just silently be killed)?

src/lib.rs
```diff
@@ -50,36 +58,95 @@ pub fn run(args: Args) {
...
+                            return WalkState::Continue;
+                        }
+                        sender.send(entry).unwrap();
+                        WalkState::Continue
+                    }
+                })
+            });
+        });
+        Self {
+            receiver_iterator: receiver.into_iter(),
```
peterstuart, line 129: Doesn't really matter, but why convert this into an `iter` here rather than in `next` below? Because then you'd have to consume it?

src/lib.rs
peterstuart: Yeah, your reasoning makes sense to me there. If you're confident that the thread will be cleanup up (either by finishing, or the parent process terminating), I don't really see a reason to join it.

peterstuart (approved):
I didn't actually run it, but this reads really nicely to me, and feels more "understandable", for whatever reason, than the previous version.

src/lib.rs
helixbass: Ya I guess (I was confused `Receiver` implements `IntoIterator` not `Iterator` directly) it seemed like what this struct needed as its internal state was the actual "inner `Iterator`" (which its `Iterator` implementation can just literally delegate to, interesting how unless there's a reason to hold onto the thread join handle reference (which I'm not sure - do bad things happen if you don't keep a thread join handle in scope?) this struct boils down to a basically non-existent wrapper around that inner `Iterator`)

src/lib.rs
helixbass: Ok per other comment then reduced this down to just a "pure delegation" to `self.receiver_iterator.next()`

PR #17, opened 2023-06-07
One passing test

helixbass:
In this PR:
- first pass at a pattern for testing the actual command-line matches output for a given (inline) query

To test:
Running `cargo test` you should see a passing test (and if you eg comment out one of the `ExpectedMatch`'s for that test then the test should fail)

### DIFF
--- a/tests/output.rs
+++ b/tests/output.rs
@@ -1,3 +1,4 @@
+#![allow(clippy::into_iter_on_ref)]
 use std::{
     env,
     ffi::OsStr,
@@ -63,7 +64,7 @@ impl ExpectedMatch {
     }
 }
 
-const MATCH_OUTPUT_LINE_REGEX_STR: &'static str = r#"(^|\n).+:\d+:"#;
+const MATCH_OUTPUT_LINE_REGEX_STR: &str = r#"(^|\n).+:\d+:"#;
 
 fn predicate_from_expected_matches(expected_matches: &[ExpectedMatch]) -> BoxPredicate<str> {
     expected_matches.into_iter().fold(
@@ -134,7 +135,7 @@ fn get_fixture_dir_path_from_name(fixture_dir_name: &str) -> PathBuf {
     path
 }
 
-const FUNCTION_ITEM_QUERY_SOURCE: &'static str = "(function_item) @function_item";
+const FUNCTION_ITEM_QUERY_SOURCE: &str = "(function_item) @function_item";
 
 static RUST_PROJECT_FUNCTION_ITEM_EXPECTED_MATCHES: Lazy<FixtureQueryExpectedMatches> =
     Lazy::new(|| FixtureQueryExpectedMatches {
--- a/tests/output.rs
+++ b/tests/output.rs
@@ -1,12 +1,13 @@
-use assert_cmd::prelude::*;
-use once_cell::sync::Lazy;
-use predicates::{prelude::*, BoxPredicate};
 use std::{
     env,
     ffi::OsStr,
     path::{Path, PathBuf},
     process::Command,
 };
+
+use assert_cmd::prelude::*;
+use once_cell::sync::Lazy;
+use predicates::{prelude::*, BoxPredicate};
 
 struct Location {
     line: usize,
--- a/tests/output.rs
+++ b/tests/output.rs
@@ -1,0 +1,174 @@
+use assert_cmd::prelude::*;
+use once_cell::sync::Lazy;
+use predicates::{prelude::*, BoxPredicate};
+use std::{
+    env,
+    ffi::OsStr,
+    path::{Path, PathBuf},
+    process::Command,
+};
+
+struct Location {
+    line: usize,
+    #[allow(dead_code)]
+    column: usize,
+}
+
+impl From<(usize, usize)> for Location {
+    fn from(value: (usize, usize)) -> Self {
+        Self {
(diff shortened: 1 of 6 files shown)


tests/output.rs
```diff
@@ -0,0 +1,174 @@
...
+            .map(|(line_index, line)| {
+                format!(
+                    "{}:{}:{}",
+                    self.relative_file_path,
+                    self.range.start.line + line_index + 1,
+                    line
+                )
+            })
+            .collect::<Vec<_>>()
+            .join("\n")
```
helixbass, line 61: Since the current default(/only) output format shows all of the lines corresponding to the captured AST node, we're concatenating multiple formatted output lines per expected match here

tests/output.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+const MATCH_OUTPUT_LINE_REGEX_STR: &'static str = r#"(^|\n).+:\d+:"#;
+
+fn predicate_from_expected_matches(expected_matches: &[ExpectedMatch]) -> BoxPredicate<str> {
+    expected_matches.into_iter().fold(
+        BoxPredicate::new(
+            predicate::str::is_match(MATCH_OUTPUT_LINE_REGEX_STR)
+                .unwrap()
+                .count(
```
helixbass, line 72: This is asserting that there aren't _more_ matches than what we specified (via a hacky regex for detecting a formatted match)

tests/output.rs
```diff
@@ -0,0 +1,174 @@
...
+            .join("\n")
+    }
+}
+
+const MATCH_OUTPUT_LINE_REGEX_STR: &'static str = r#"(^|\n).+:\d+:"#;
+
+fn predicate_from_expected_matches(expected_matches: &[ExpectedMatch]) -> BoxPredicate<str> {
+    expected_matches.into_iter().fold(
+        BoxPredicate::new(
+            predicate::str::is_match(MATCH_OUTPUT_LINE_REGEX_STR)
```
helixbass, line 70: This approach using [`predicates`](https://docs.rs/predicates/latest/predicates/) is from [this guide](https://rust-cli.github.io/book/tutorial/testing.html#testing-cli-applications-by-running-them), which seems not-insane for a first pass

tests/output.rs
```diff
@@ -0,0 +1,174 @@
...
+    let root_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+    let mut path: PathBuf = root_dir.into();
+    path.push("tests/fixtures");
+    path.push(fixture_dir_name);
+    path
+}
+
+const FUNCTION_ITEM_QUERY_SOURCE: &'static str = "(function_item) @function_item";
+
+static RUST_PROJECT_FUNCTION_ITEM_EXPECTED_MATCHES: Lazy<FixtureQueryExpectedMatches> =
```
helixbass, line 138: I guess I'm picturing these being reusable for testing eg different output formats?

tests/output.rs
```diff
@@ -0,0 +1,174 @@
...
+            ExpectedMatch {
+                relative_file_path: "./src/lib.rs".to_owned(),
+                lines: vec![
+                    "    fn it_works() {".to_owned(),
+                    "        let result = add(2, 2);".to_owned(),
+                    "        assert_eq!(result, 4);".to_owned(),
+                    "    }".to_owned(),
+                ],
+                range: ((11, 0), (14, 4)).into(),
+            },
```
peterstuart, line 167: This reads nicely, but if you wanted to get really crazy/cute with it, you could write a little test parser a-la-rust-analyzer tests? I could imagine writing tests like this?

```rs
#[test]
fn test_something() {
assert_output!(r#"
$ tree-sitter-grep --some-option
.src/helpers.rs:10:   pub fn helper() {}
...
"#);
```

Where you write `$ the command to actually execute` followed by the expected output as string.

(I don't know that it's actually a good idea, just something that occurred to me.)

peterstuart: Snapshot tests also seems appealing for something like this.

peterstuart (approved):

tests/output.rs
peterstuart: Ah yeah, I'd also been assuming without looking into it that the ordering was not deterministic. I think the snapshotter could sort the output, though, and then it'd be back to deterministic?

> so does it not do something similar to what we're doing and then also do the matching in parallel (without then collecting/sorting the aggregated matches)?

Huh, yeah I'd have assumed the matching was also in parallel.

tests/output.rs
```diff
@@ -0,0 +1,176 @@
+#![allow(clippy::into_iter_on_ref)]
```
helixbass, line 1: I disagree stylistically with this lint, I think using `.into_iter()` on a reference type is fine and it is a stunted mental model to think that `.into_iter()` means "it moves the contents of the collection into the iterator" (if that's what the lint [docs](https://rust-lang.github.io/rust-clippy/master/index.html#/into_iter_on_ref) are indeed telling me). No, a reference type obviously doesn't own the contents of the collection, and it has a different `.into_iter()` implementation. Encouraging people to "conflate" `.into_iter()` with "owning" seems like a terrible idea

Also I think it's insane that there's no such thing as eg a workspace-wide config file for configuring individual Clippy lints. Like here I'd have to add this separately to `src/lib.rs` if I wanted it to "kick in" there

PR #18, opened 2023-06-08
Vimgrep mode

helixbass:
In this PR:
- add support for `--vimgrep` command-line option

Not in this PR:
- I believe this should set me up to more easily write some sort of (neo)vim plugin wrapper to eg show matches in the quickfix list but I haven't tried that/am not super familiar with what's involved in doing that

To test:
If you run any previously-working command-line with `--vimgrep` added, you should see matches formatted like:
```
path/without/leading/dot/slash.rs:line:column:line contents
```
(where if the matched AST node spans multiple lines only the first line is included in the formatted output)

`cargo test` should pass

Based on `test-matches`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,8 +7,6 @@ edition = "2021"
 
 [dependencies]
 clap = { version = "4.3.0", features = ["derive"] }
-# grep = "0.2.12"
-# grep = { path = "../ripgrep/crates/grep" }
 grep = { git = "https://github.com/helixbass/ripgrep", rev = "b83cc3c" }
 ignore = "0.4.20"
 libc = "0.2.144"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,13 +7,16 @@ edition = "2021"
 
 [dependencies]
 clap = { version = "4.3.0", features = ["derive"] }
-grep = "0.2.12"
+# grep = "0.2.12"
+# grep = { path = "../ripgrep/crates/grep" }
+grep = { git = "https://github.com/helixbass/ripgrep", rev = "b83cc3c" }
 ignore = "0.4.20"
 libc = "0.2.144"
 libloading = "0.8.0"
 once_cell = "1.17.1"
 rayon = "1.7.0"
 regex = "1.8.2"
+termcolor = "1.2.0"
 tree-sitter = "0.20.10"
 tree-sitter-rust = "0.20.3"
 tree-sitter-typescript = "0.20.2"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,6 +3,7 @@ use grep::matcher::Match;
 use grep::matcher::Matcher;
 use grep::matcher::NoCaptures;
 use grep::matcher::NoError;
+use grep::searcher::Searcher;
 use grep::searcher::SearcherBuilder;
 use ignore::WalkParallel;
 use ignore::WalkState;
@@ -41,6 +42,8 @@ pub struct Args {
     pub filter: Option<String>,
     #[arg(short = 'a', long)]
     pub filter_arg: Option<String>,
+    #[arg(long)]
+    pub vimgrep: bool,
 }
 
 #[derive(clap::Args)]
@@ -49,6 +52,20 @@ pub struct QueryArgs {
     pub path_to_query_file: Option<PathBuf>,
     #[arg(short, long)]
     pub query_source: Option<String>,
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+enum OutputMode {
+    Normal,
+    Vimgrep,
+}
+
+fn get_output_mode(args: &Args) -> OutputMode {
+    if args.vimgrep {
+        OutputMode::Vimgrep
+    } else {
+        OutputMode::Normal
+    }
 }
 
 pub fn run(args: Args) {
@@ -64,11 +81,12 @@ pub fn run(args: Args) {
             .capture_index_for_name(capture_name)
             .expect(&format!("Unknown capture name: `{}`", capture_name))
     });
+    let output_mode = get_output_mode(&args);
 
     get_project_file_walker(&*supported_language)
         .into_parallel_iterator()
         .for_each(|project_file_dir_entry| {
-            let mut printer = grep::printer::Standard::new_no_color(io::stdout());
+            let mut printer = get_printer(output_mode);
             let path = project_file_dir_entry.path();
 
             let matcher = TreeSitterMatcher::new(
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -7,13 +7,14 @@ edition = "2021"
...
 clap = { version = "4.3.0", features = ["derive"] }
-grep = "0.2.12"
+grep = { git = "https://github.com/helixbass/ripgrep", rev = "b83cc3c" }
 ignore = "0.4.20"
 libc = "0.2.144"
 libloading = "0.8.0"
 once_cell = "1.17.1"
 rayon = "1.7.0"
 regex = "1.8.2"
+termcolor = "1.2.0"
```
helixbass, line 17: When I had to write out the type of a `Searcher` (I think?) I needed to reference a type from `termcolor` which I guess is another one of the "`ripgrep`-derived" crates

src/lib.rs
```diff
@@ -64,12 +82,13 @@ pub fn run(args: Args) {
...
     });
+    let output_mode = get_output_mode(&args);
 
     get_project_file_walker(&*supported_language)
         .into_parallel_iterator()
         .for_each(|project_file_dir_entry| {
-            let mut printer = grep::printer::Standard::new_no_color(io::stdout());
-            let path = project_file_dir_entry.path();
+            let mut printer = get_printer(output_mode);
+            let path = format_relative_path(project_file_dir_entry.path(), output_mode);
```
helixbass, line 91: It seems like `.search_path()`/`printer.sink_with_path()` will literally print whatever relative path formatting is in the `Path` that you pass it, I guess the `DirEntry.path()` here includes a leading `./` and that's what we wanted for the existing output format but `--vimgrep` mode appears to strip the leading `./` so I just massage it here based on the output mode (before then passing to `.search_path()`/`printer.sink_with_path()`)

src/lib.rs
```diff
@@ -79,14 +98,43 @@ pub fn run(args: Args) {
...
-                .build()
+            get_searcher(output_mode)
                 .search_path(&matcher, path, printer.sink_with_path(&matcher, path))
                 .unwrap();
         });
 }
 
+fn format_relative_path(path: &Path, output_mode: OutputMode) -> &Path {
+    if output_mode == OutputMode::Vimgrep && path.starts_with("./") {
+        path.strip_prefix("./").unwrap()
```
helixbass, line 109: I have no idea if there are higher-level/more canonical ways to "normalize" relative paths (or what that even means)?

src/lib.rs
```diff
@@ -79,14 +98,43 @@ pub fn run(args: Args) {
...
+    } else {
+        path
+    }
+}
+
+fn get_printer(output_mode: OutputMode) -> grep::printer::Standard<termcolor::NoColor<io::Stdout>> {
+    match output_mode {
+        OutputMode::Normal => grep::printer::Standard::new_no_color(io::stdout()),
+        OutputMode::Vimgrep => grep::printer::StandardBuilder::new()
+            .per_match(true)
```
helixbass, line 119: These options (and the additional option to the `SearcherBuilder` below) were from looking for references to the `vimgrep` option in the `ripgrep` source code, some of them may already be the default value

src/lib.rs
```diff
@@ -186,12 +234,6 @@ impl Matcher for TreeSitterMatcher<'_> {
 
     fn find_at(&self, haystack: &[u8], at: usize) -> Result<Option<Match>, Self::Error> {
         let mut matches_info = self.matches_info.borrow_mut();
-        if let Some(matches_info) = matches_info.as_ref() {
```
helixbass, line 189: We don't in fact always get called with progressively shorter trailing substrings of the file contents (per the reason `grep` was forked there are other calls for doing some kind of "re-matching" happening)

src/lib.rs
```diff
@@ -208,15 +250,7 @@ impl Matcher for TreeSitterMatcher<'_> {
                 text_len: haystack.len(),
             }
         });
-        if matches_info.matches.is_empty() {
-            return Ok(None);
-        }
-        let match_ = matches_info.matches.remove(0);
```
helixbass, line 214: ...and so we can't just blindly "pop + return" one match at a time, we have to keep them all around and give it back the first one that starts after the (calculated via the pseudo-invariant) file contents offset of the substring that it's passing us

tests/output.rs
```diff
@@ -44,62 +44,106 @@ struct ExpectedMatch {
     pub range: Range,
 }
 
-impl ExpectedMatch {
-    pub fn expected_output_text(&self) -> String {
```
helixbass, line 48: I moved this onto a method on `OutputMode` below (which seems questionable given the number of references to fields on this struct)

tests/output.rs
```diff
@@ -44,62 +44,106 @@ struct ExpectedMatch {
...
         BoxPredicate::new(
             predicate::str::is_match(MATCH_OUTPUT_LINE_REGEX_STR)
                 .unwrap()
-                .count(
-                    expected_matches
-                        .into_iter()
-                        .map(|expected_match| expected_match.lines.len())
-                        .sum(),
-                ),
+                .count(output_mode.get_expected_total_number_of_match_lines(expected_matches)),
```
helixbass, line 57: ...ya something felt right when writing it to put this and the other method on `OutputMode` but when reading it I'm not so sure

tests/output.rs
```diff
@@ -44,62 +44,106 @@ struct ExpectedMatch {
...
         ),
         |predicate, expected_match| {
-            BoxPredicate::new(predicate.and(predicate_from_expected_match(expected_match)))
+            BoxPredicate::new(
+                predicate.and(predicate_from_expected_match(expected_match, output_mode)),
```
helixbass, line 61: This one (`predicate_from_expected_match()`) could be a method on `ExpectedMatch`?

src/lib.rs
```diff
@@ -51,6 +55,20 @@ pub struct QueryArgs {
...
+    Vimgrep,
+}
+
+fn get_output_mode(args: &Args) -> OutputMode {
+    if args.vimgrep {
+        OutputMode::Vimgrep
+    } else {
+        OutputMode::Normal
+    }
+}
```
peterstuart, line 70: I wonder if there's some fancy way you explain this to clap declaratively.

peterstuart (approved):

src/lib.rs
```diff
@@ -79,14 +98,43 @@ pub fn run(args: Args) {
...
+
+fn get_printer(output_mode: OutputMode) -> grep::printer::Standard<termcolor::NoColor<io::Stdout>> {
+    match output_mode {
+        OutputMode::Normal => grep::printer::Standard::new_no_color(io::stdout()),
+        OutputMode::Vimgrep => grep::printer::StandardBuilder::new()
+            .per_match(true)
+            .per_match_one_line(true)
+            .column(true)
+            .heading(false)
+            .path(true)
```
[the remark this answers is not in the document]
helixbass: Ok removed

PR #20, opened 2023-06-08
test sorted output

helixbass:
In this PR:
- per [this comment](https://github.com/helixbass/tree-sitter-grep/pull/17#discussion_r1222321526) switch to a simpler "parsed command + output" style for integration tests of whether matches are output as expected

To test:
Tests should still pass

Based on `vimgrep-mode`

### DIFF
--- a/tests/output.rs
+++ b/tests/output.rs
@@ -1,176 +1,13 @@
 use assert_cmd::prelude::*;
-use once_cell::sync::Lazy;
-use predicates::{prelude::*, BoxPredicate};
-use std::{
-    env,
-    ffi::OsStr,
-    path::{Path, PathBuf},
-    process::Command,
-};
+use predicates::prelude::*;
+use std::{env, path::PathBuf, process::Command};
 
-struct Location {
-    line: usize,
-    #[allow(dead_code)]
-    column: usize,
-}
-
-impl From<(usize, usize)> for Location {
-    fn from(value: (usize, usize)) -> Self {
-        Self {
-            line: value.0,
-            column: value.1,
-        }
-    }
-}
-
-struct Range {
-    pub start: Location,
-    #[allow(dead_code)]
-    pub end: Location,
-}
-
-impl From<((usize, usize), (usize, usize))> for Range {
-    fn from(value: ((usize, usize), (usize, usize))) -> Self {
-        Self {
-            start: value.0.into(),
-            end: value.1.into(),
-        }
-    }
-}
-
-struct ExpectedMatch {
-    pub relative_file_path: String,
-    pub lines: Vec<String>,
-    pub range: Range,
-}
-
-const MATCH_OUTPUT_LINE_REGEX_STR: &'static str = r#"(^|\n).+:\d+:"#;
-
-fn predicate_from_expected_matches(
-    expected_matches: &[ExpectedMatch],
-    output_mode: OutputMode,
-) -> BoxPredicate<str> {
-    expected_matches.into_iter().fold(
-        BoxPredicate::new(
-            predicate::str::is_match(MATCH_OUTPUT_LINE_REGEX_STR)
-                .unwrap()
-                .count(output_mode.get_expected_total_number_of_match_lines(expected_matches)),
-        ),
-        |predicate, expected_match| {
-            BoxPredicate::new(
-                predicate.and(predicate_from_expected_match(expected_match, output_mode)),
-            )
-        },
-    )
-}
-
-fn predicate_from_expected_match(
-    expected_match: &ExpectedMatch,
-    output_mode: OutputMode,
-) -> BoxPredicate<str> {
-    BoxPredicate::new(predicate::str::contains(
-        output_mode.expected_output_text(expected_match),
-    ))
-}
-
-fn assert_expected_matches<TArg: AsRef<OsStr>>(
-    args: impl IntoIterator<Item = TArg>,
-    current_dir: impl AsRef<Path>,
-    expected_matches: &[ExpectedMatch],
-    output_mode: OutputMode,
-) {
-    Command::cargo_bin("tree-sitter-grep")
-        .unwrap()
-        .args(args)
-        .current_dir(current_dir)
-        .assert()
-        .success()
-        .stdout(predicate_from_expected_matches(
-            expected_matches,
-            output_mode,
-        ));
-}
-
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-enum OutputMode {
-    Normal,
-    Vimgrep,
-}
-
-impl OutputMode {
-    pub fn get_expected_total_number_of_match_lines(
-        &self,
-        expected_matches: &[ExpectedMatch],
-    ) -> usize {
-        match self {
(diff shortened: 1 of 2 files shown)


tests/output.rs
```diff
@@ -1,228 +1,118 @@
...
-            column: value.1,
-        }
-    }
-}
-
-struct Range {
-    pub start: Location,
-    #[allow(dead_code)]
-    pub end: Location,
+fn get_fixture_dir_path_from_name(fixture_dir_name: &str) -> PathBuf {
```
helixbass, line 13: Existing, just moved

tests/output.rs
```diff
@@ -1,228 +1,118 @@
...
-    pub fixture_dir_name: String,
-    pub language: String,
-    pub query_source: String,
-    pub expected_matches: Vec<ExpectedMatch>,
+fn do_sorted_lines_match(actual_output: &str, expected_output: &str) -> bool {
+    let mut actual_lines = actual_output.split("\n").collect::<Vec<_>>();
+    actual_lines.sort();
+    let mut expected_lines = expected_output.split("\n").collect::<Vec<_>>();
+    expected_lines.sort();
+    actual_lines == expected_lines
```
helixbass, line 86: This isn't comparing the lines in some "canonical sensible" order, by doing lexicographic sorting here you'd get things like a match on line 12 coming before a match on line 3 etc, but if all of the individual sorted lines match we're probably in good shape unless we're worried about eg multi-line matches not being output in the correct order wrt each other?

tests/output.rs
```diff
@@ -1,228 +1,118 @@
...
-        ],
-    });
-
 #[test]
 fn test_query_inline() {
-    RUST_PROJECT_FUNCTION_ITEM_EXPECTED_MATCHES.assert_expected_matches(OutputMode::Normal);
+    assert_sorted_output(
+        "rust_project",
+        r#"
+            $ tree-sitter-grep --query-source '(function_item) @function_item' --language rust
```
helixbass, line 94: I think this is great, the pattern I'd set up was feeling clunky/not exactly paying huge dividends for "reuse" so figured I'd switch to this

tests/output.rs
```diff
@@ -1,228 +1,118 @@
...
+use std::{env, path::PathBuf, process::Command};
 
-struct Location {
-    line: usize,
-    #[allow(dead_code)]
-    column: usize,
+#[macro_export]
+macro_rules! regex {
+    ($re:literal $(,)?) => {{
+        static RE: once_cell::sync::OnceCell<regex::Regex> = once_cell::sync::OnceCell::new();
```
peterstuart, line 8: Unrelated, but I think that part of `once_cell` was stabilized in the latest version of Rust, so you might not need the dependency here any more.

tests/output.rs
```diff
@@ -1,228 +1,118 @@
...
 }
 
-impl From<((usize, usize), (usize, usize))> for Range {
-    fn from(value: ((usize, usize), (usize, usize))) -> Self {
-        Self {
-            start: value.0.into(),
-            end: value.1.into(),
+fn parse_command_and_output(command_and_output: &str) -> CommandAndOutput {
+    let mut lines = command_and_output.split("\n").collect::<Vec<_>>();
+    if !lines.is_empty() {
```
peterstuart, line 24: What would it mean if `lines` *was* empty here? Should it crash/fail the test?

tests/output.rs
```diff
@@ -1,228 +1,118 @@
...
+        if regex!(r#"^\s*$"#).is_match(&lines[0]) {
```
peterstuart, line 25: This is fine, but personally I have to think a bit to mentally parse the regex, so I'd probably write:

```rs
if lines[0].trim().is_empty() {
```

tests/output.rs
```diff
@@ -1,228 +1,118 @@
...
+            lines.pop();
+        }
+    }
+    let output: String = lines
+        .into_iter()
+        .map(|line| {
+            assert!(line.starts_with(indent));
+            format!("{}\n", strip_indent(line, indent))
+        })
+        .collect();
```
peterstuart, line 43: Unless the whitespace is meaningful in some cases, I think it might be good enough (and simpler) to just `.trim()` or `.trim_start()` all the lines instead of making sure the indentation matches.

tests/output.rs
peterstuart: > unless we're worried about eg multi-line matches not being output in the correct order wrt each other

Like if there were overlapping matches, you might get 2 different matches from the same line in the same file? (Which would be weird to sort if they're multi-line matches?)

This seems good enough ™️ to me for now.

tests/output.rs
```diff
@@ -1,228 +1,118 @@
...
 }
 
-fn assert_expected_matches<TArg: AsRef<OsStr>>(
-    args: impl IntoIterator<Item = TArg>,
-    current_dir: impl AsRef<Path>,
-    expected_matches: &[ExpectedMatch],
-    output_mode: OutputMode,
-) {
-    Command::cargo_bin("tree-sitter-grep")
+fn assert_sorted_output(fixture_dir_name: &str, command_and_output: &str) {
```
peterstuart, line 64: The way you have tests written currently (essentially 1-statement per test) it doesn't matter, but I think it's generally worth making assert functions macros, so that you get the panic reported for a line that's actually in a test.

peterstuart (approved):

tests/output.rs
peterstuart: Also maybe worth considering using something like https://crates.io/crates/pretty_assertions to get nice diffs in the assertions.

helixbass: > I think it's generally worth making assert functions macros, so that you get the panic reported for a line that's actually in a test

Ah interesting. I wonder if `#[track_caller]` would accomplish the same thing (that's what I saw being used in `moxie` for the stuff related to "what's the current source-code location", I think it basically says "if I get asked what my source-code location is (eg for backtraces?) I'll lie and tell them the location of whoever called me instead")?

> Also maybe worth considering using something like https://crates.io/crates/pretty_assertions to get nice diffs in the assertions

Ya that would be nice, and there have been a couple places I might've reached for [`speculoos`](https://crates.io/crates/speculoos)-style assertions (which I've been liking for the Typescript compiler in Rust tests), but I'm not sure I totally understand how/where to integrate that with these existing assertions eg `predicates` that that guide was using

tests/output.rs
helixbass: Ya technically my understanding would be "the only way this would give a false positive (where "positive" == "passed") is if what was actually output was the same exact set of lines just in a different order (but where that different order wasn't acceptable ie wasn't just a different output order of the matches)", which seems like the only surface area there is "some multi-line match wasn't printed out contiguously"

I agree it seems like significantly more work to try and test that and I'd guess we either have interleaving problems or we don't

tests/output.rs
peterstuart: With the current setup, I think CI is running whatever the latest Rust is at the time. If this becomes a library crate, I guess we'd probably want to have CI run against ever supported Rust version, or at least the oldest and newest?

PR #21, opened 2023-06-09
Specify files

### LINKED ISSUE #16: Support specifying which files to match against
Maybe try and mimic the CLI of existing things like `grep`/`ripgrep`?

helixbass:
In this PR:
- (optionally) specify which files to search on the command line
- update omission of `./` in formatted match output paths to not be based on "output mode"
- update searcher/printer "architecture" to still parallelize searching/printing but hopefully not be subject to "interleaving" of printed matches

To test:
If you don't explicitly specify any files on the command line it should still search "all" files
If you specify one or more file paths as positional command-line arguments it should only search/show matches for those files
Regardless of output "mode" (ie presence or absence of `--vimgrep`) it should only include `./` in printed match output paths if you explicitly specified a corresponding file path that also included a `./`
You shouldn't see interleaved results

Performance-wise (based again on simple testing via `time`) it looks like this (presumably the searcher/printer architecture change) slowed it down just a bit, maybe from ~280-300ms -> ~290-320ms (for the same test case I was using before)

Closes #16 

Based on `test-sorted-output`


src/lib.rs
```diff
@@ -47,9 +54,24 @@ pub struct Args {
     pub vimgrep: bool,
 }
 
+impl Args {
+    pub fn use_paths(&self) -> Vec<PathBuf> {
```
helixbass, line 58: This mimics what I saw in `ripgrep`, I think I misunderstood the `WalkBuilder::new()` argument and thought it was for the "current directory" but no `ripgrep` just passes it the first of its paths (which like this it ensures at least includes `./` as a default) and then explicitly `.add()`'s the rest of those paths

src/lib.rs
```diff
@@ -47,9 +54,24 @@ pub struct Args {
...
+
+    pub fn is_using_default_paths(&self) -> bool {
+        self.paths.is_empty()
+    }
+}
+
 #[derive(clap::Args)]
 #[group(required = true, multiple = false)]
 pub struct QueryArgs {
+    #[arg(short = 'Q', long = "query-file")]
```
helixbass, line 74: The path to a query file had been a positional command-line argument, I'm guessing it's uncontroversial that we should mimic grep-like tools and have paths be the positional command-line arguments?

src/lib.rs
```diff
@@ -83,12 +105,15 @@ pub fn run(args: Args) {
             .expect(&format!("Unknown capture name: `{}`", capture_name))
     });
     let output_mode = get_output_mode(&args);
+    let buffer_writer = BufferWriter::stdout(ColorChoice::Never);
```
helixbass, line 108: Ok this stuff is for avoiding the interleaving printed matches (which I started seeing consistently once I added another searched file to the `rust_project/` test fixture directory)

[This method](https://github.com/BurntSushi/ripgrep/blob/master/crates/core/main.rs#L127) is where you can see how `ripgrep` is doing it, and I just mimicked what seemed to be the relevant parts (with some changes to account for the fact that we're doing separate multi-threading for running the queries (via `rayon`) vs they are just using the `WalkParallel.run()` callback for multi-threading)

Specifically here a `BufferWriter` I believe is responsible for avoiding interleaving

And so we spawn/cache a printer + searcher per thread and the printer leverages the `BufferWriter` or something like that

src/lib.rs
```diff
@@ -83,12 +105,15 @@ pub fn run(args: Args) {
...
-    get_project_file_walker(&*supported_language)
+    get_project_file_walker(&*supported_language, &args.use_paths())
         .into_parallel_iterator()
         .for_each(|project_file_dir_entry| {
-            let mut printer = get_printer(output_mode);
-            let path = format_relative_path(project_file_dir_entry.path(), output_mode);
+            let printer = get_printer(&buffer_writer, output_mode);
+            let mut printer = printer.borrow_mut();
+            let path =
+                format_relative_path(project_file_dir_entry.path(), args.is_using_default_paths());
```
helixbass, line 116: Here we make the presence of leading `./` a function of whether we're in "default paths" mode or not (per `ripgrep`'s apparent behavior)

src/lib.rs
```diff
@@ -99,33 +124,67 @@ pub fn run(args: Args) {
...
-fn format_relative_path(path: &Path, output_mode: OutputMode) -> &Path {
-    if output_mode == OutputMode::Vimgrep && path.starts_with("./") {
-        path.strip_prefix("./").unwrap()
-    } else {
-        path
-    }
+type Printer = Standard<Buffer>;
+
+thread_local! {
+    static PRINTER: OnceCell<(Rc<RefCell<Printer>>, OutputMode)> = Default::default();
```
helixbass, line 137: Given wanting to (mimiicking my understanding of what `ripgrep` was doing via the `WalkParallel.run()` callback) have a single printer/searcher per thread, I reached for a `thread_local!` + `OnceCell` + `Rc`-wrapper (since a reference to the `thread_local!` value can't escape its `.with()` callback, so you'd either have to re-arrange the code using the printer to be expressed as inside such a callback or else you can use eg an `Rc`-wrapper to let a reference "leak" outside of the `.with()` callback) + `RefCell` inner wrapper (since we need to call `&mut self` methods on the printer/searcher, and inside an `Rc` you need interior mutability in order to do that)

src/lib.rs
```diff
@@ -83,12 +105,15 @@ pub fn run(args: Args) {
...
+    let buffer_writer = BufferWriter::stdout(ColorChoice::Never);
 
-    get_project_file_walker(&*supported_language)
+    get_project_file_walker(&*supported_language, &args.use_paths())
         .into_parallel_iterator()
         .for_each(|project_file_dir_entry| {
-            let mut printer = get_printer(output_mode);
-            let path = format_relative_path(project_file_dir_entry.path(), output_mode);
+            let printer = get_printer(&buffer_writer, output_mode);
+            let mut printer = printer.borrow_mut();
```
helixbass, line 114: (See below comment for why this is in a `RefCell` in the first place) Specifically here my understanding of why this should never result in a runtime borrow error is that (like all `RefCell`'s) this `RefCell` is local to the current thread so as long as we drop it when we're done with it (and don't eg call back into this code path in a nested way) then there should be no way for a simultaneous borrow to occur

src/lib.rs
```diff
@@ -99,33 +124,67 @@ pub fn run(args: Args) {
             );
 
             get_searcher(output_mode)
+                .borrow_mut()
                 .search_path(&matcher, path, printer.sink_with_path(&matcher, path))
                 .unwrap();
+            buffer_writer.print(printer.get_mut()).unwrap();
```
helixbass, line 130: I think what's happening here (+ in `get_printer()`) is that the `BufferWriter` hands out `Buffer`'s (which serve as each printer's inner `Writer`) but then here there's just the single `BufferWriter` (across all threads) and it's getting passed a reference to this printer's `Buffer` (via `.get_mut()`) in order to then actually print out its contents?

src/lib.rs
```diff
@@ -99,33 +124,67 @@ pub fn run(args: Args) {
...
+fn get_printer(buffer_writer: &BufferWriter, output_mode: OutputMode) -> Rc<RefCell<Printer>> {
+    PRINTER.with(|printer| {
+        let (printer, output_mode_when_initialized) = printer.get_or_init(|| {
+            (
+                Rc::new(RefCell::new(create_printer(buffer_writer, output_mode))),
+                output_mode,
+            )
+        });
+        assert!(
+            *output_mode_when_initialized == output_mode,
```
helixbass, line 148: Just an "invariant assertion" related to the fact that we're "caching" these printers/searchers but that they're in fact contingent (at the time of creation/caching) on an `OutputMode`

src/lib.rs
```diff
@@ -99,33 +124,67 @@ pub fn run(args: Args) {
...
+            "Using multiple output modes not supported"
+        );
+        printer.clone()
+    })
 }
 
-fn get_printer(output_mode: OutputMode) -> grep::printer::Standard<termcolor::NoColor<io::Stdout>> {
+fn create_printer(buffer_writer: &BufferWriter, output_mode: OutputMode) -> Printer {
     match output_mode {
-        OutputMode::Normal => grep::printer::Standard::new_no_color(io::stdout()),
```
helixbass, line 117: I got rid of the `*_no_color()` because it seemed like that's actually the default and it was making `printer.get_mut()` (above) not directly return a `Buffer` (because specifically I think `*_no_color()` would return a `Standard<NoColor<Buffer>>` instead of just a `Standard<Buffer>`)

tests/fixtures/rust_project/src/lib.rs
```diff
@@ -14,3 +14,5 @@ mod tests {
         assert_eq!(result, 4);
     }
 }
+
+mod stop;
```
helixbass, line 18: I put this at the bottom to avoid messing up existing match line numbers

Really this project doesn't need to be valid/compile I suppose but whatever

src/lib.rs
```diff
@@ -47,9 +54,24 @@ pub struct Args {
     pub vimgrep: bool,
 }
 
+impl Args {
+    pub fn use_paths(&self) -> Vec<PathBuf> {
+        if self.paths.is_empty() {
+            vec![Path::new("./").to_owned()]
```
peterstuart, line 60: Not that you probably care right now, but do you know if this works on Windows?

src/lib.rs
peterstuart: Yeah that seems right to me.

tests/fixtures/rust_project/function-item.scm
```diff
@@ -0,0 +1 @@
+(function_item) @function_item
```
peterstuart, line 1: I meant to ask in some other PR, what's up with the "scm" extension on these files?

tests/output.rs
```diff
@@ -105,14 +211,19 @@ fn test_query_inline() {
...
+            $ tree-sitter-grep --query-source '(function_item) @function_item' --language rust src/lib.rs ./src/helpers.rs
+            ./src/helpers.rs:1:pub fn helper() {}
+            src/lib.rs:3:pub fn add(left: usize, right: usize) -> usize {
+            src/lib.rs:4:    left + right
+            src/lib.rs:5:}
+            src/lib.rs:12:    fn it_works() {
+            src/lib.rs:13:        let result = add(2, 2);
+            src/lib.rs:14:        assert_eq!(result, 4);
+            src/lib.rs:15:    }
+        "#,
```
peterstuart, line 227: Refreshing my memory, if I passed in `src/*.rs` to tree-sitter-grep, that'll actually get expanded by my shell, right? So TSG will actually be passed a list of `.rs` files?

peterstuart (approved):

src/lib.rs
helixbass: Ya that had crossed my mind, I'd guess there's some "path"-y stuff that currently wouldn't work but no idea, opened an issue

I should get a Windows laptop that I can dual-boot Linux on since I've been wanting a way to run Linux locally too

tests/fixtures/rust_project/function-item.scm
helixbass: [This](https://tree-sitter.github.io/tree-sitter/syntax-highlighting#queries) explains it, I guess "Scheme" or (perhaps after-the-fact) "source code matching"

tests/output.rs
helixbass: Yes I was thinking about that too, I believe that is accurate. I tried adding one test with such a glob but something was failing (though the equivalent "actual" command-line was working as expected), maybe eg `shlex` got confused?

It looks like `ripgrep` also supports passing it "globs" - I think how this works (eg you can do this with ESLint or Prettier I remember) is that if you put it in single quotes eg `'**/*.rs'` then the shell doesn't expand it for you

But so that presumably is not currently supported. Eventually it would probably make sense to basically go through all of the eg `ripgrep` command-line options and see which ones also seem like they'd be appropriate/relevant and mimic/support them?

src/lib.rs
```diff
@@ -92,32 +115,67 @@ pub fn run(args: Args) {
                 args.filter_arg.clone(),
             );
 
+            printer.get_mut().clear();
```
helixbass, line 118: This is what was missing (vs what `ripgrep` [does](https://github.com/BurntSushi/ripgrep/blob/4fcb1b2202b97c5a21894672232700225223a138/crates/core/main.rs#L159))

PR #24, opened 2023-06-10
Auto-language detection

### LINKED ISSUE #22: "Auto-detect" language?
It seems possibly desirable to not have to specify the `--language` command line argument every time

Biggest "issue" would seem to be that the tree-sitter query presumably would only parse for the actual intended target language (maybe others?)

So maybe we know how to map from filename suffix to supported language (which knows how to map to eg the relevant tree-sitter grammar/parser) and as `ignore` is iterating over project files we cache whether we've already tried to parse the query for that recognized language and (a) if it failed to parse we skip the file (b) if it's cached then what we cached is the parsed query so we run that (c) if we haven't tried yet for that language then we try and parse with the relevant tree-sitter parser and cache the result/failure

Not sure if this plays weirdly with the "filter plugin" API at all? Are there language-to-language differences in the "raw" tree-sitter API that would make the filter plugin choke in weird ways? But (if so) maybe that's just on you if you decide you want to omit `--language` + use a filter plugin?

helixbass:
In this PR:
- "auto-detect" language (basically iterate project files of all recognized types and try parsing them against the query and don't hard-fail if some of the languages can't parse that query)
- add tree-sitter language/parser for Javascript
- print legitimate errors for a few failure cases

To test:
Existing behavior should continue to work
If you try omitting the `--language` command-line option in a project with only a single supported file-type (eg a Rust project) then you should see the same matches as with `--language rust`. And hopefully performance (of a release build) is relatively comparable (it looked a bit slower when running against the Typescript compiler in Rust codebase but when running against a smaller Typescript codebase it seemed indistinguishable, I'm not going to worry about it too much for now)
If you try running in a directory with multiple types of supported files (Rust/Typescript/JS) (eg the newly added `tests/fixtures/mixed_project/` directory) with a query that should only be parseable by one of those language's tree-sitter grammars then you should see the expected results for those files but not see any errors related to trying to parse for the other encountered supported file types
If you try running in a directory with multiple types of supported files with a query that should be parseable by _more than one_ of those language's tree-sitter grammars (eg `(arrow_function) @a` for Typescript/JS) then you should see the expected results for files of all of those file types

Closes #22 

Based on `specify-files`


src/language.rs
```diff
@@ -44,11 +58,67 @@ impl SupportedLanguage for SupportedLanguageTypescript {
...
+    fn extensions(&self) -> Vec<&'static str> {
+        vec!["js", "jsx", "vue", "cjs", "mjs"]
+    }
+}
+
+pub fn get_javascript_language() -> SupportedLanguageJavascript {
+    SupportedLanguageJavascript
+}
+
+pub fn get_all_supported_languages() -> HashMap<SupportedLanguageName, Box<dyn SupportedLanguage>> {
```
helixbass, line 102: I started to think about making this eg a `static` `Lazy` singleton and then decided I was being premature optimization-y

src/language.rs
```diff
@@ -44,11 +58,67 @@ impl SupportedLanguage for SupportedLanguageTypescript {
...
+            Box::new(get_typescript_language()) as Box<dyn SupportedLanguage>,
+        ),
+        (
+            SupportedLanguageName::Javascript,
+            Box::new(get_javascript_language()) as Box<dyn SupportedLanguage>,
+        ),
+    ])
+}
+
+pub fn maybe_supported_language_from_path(path: &Path) -> Option<Box<dyn SupportedLanguage>> {
```
helixbass, line 119: This is the core thing mapping project file path back to supported language, so this is getting called once per project file

src/lib.rs
```diff
@@ -91,25 +99,77 @@ fn get_output_mode(args: &Args) -> OutputMode {
     }
 }
 
+struct MaybeInitializedCaptureIndex(AtomicU32);
```
helixbass, line 102: This is maybe questionable

The weird thing this is trying to address is that we can't validate a provided `--capture` (capture name) command-line argument (and sort of "parse-don't-validate"/resolve it to the corresponding capture index) until we have a successfully parsed tree-sitter query. But now that happens "on-the-fly" in parallel-world

So it seems like we want a shared atomic-y lazy-initialized resolved capture index (or the failure case also has to be considered, where they provided an invalid capture name and we'd like that to just fail "once" with a nicely printed error message)?

src/lib.rs
```diff
@@ -91,25 +99,77 @@ fn get_output_mode(args: &Args) -> OutputMode {
...
+
+impl MaybeInitializedCaptureIndex {
+    fn mark_failed(&self) {
+        self.0.store(u32::MAX - 1, Ordering::Relaxed);
```
helixbass, line 106: What I did (I guess in order to avoid having multiple atomics going on at which point the whole `Ordering` thing gets harder to reason about) was just have a single `AtomicUsize` and encode the "uninitialized" and "invalid capture name" states into its value

src/lib.rs
```diff
@@ -91,25 +99,77 @@ fn get_output_mode(args: &Args) -> OutputMode {
...
+        if let Some(already_initialized) = self.get()? {
+            return Ok(already_initialized);
+        }
+        let capture_index = match capture_name {
+            None => 0,
+            Some(capture_name) => {
+                let capture_index = query.capture_index_for_name(capture_name);
+                if capture_index.is_none() {
+                    self.mark_failed();
+                    fail(&format!("invalid capture name '{}'", capture_name));
```
helixbass, line 128: So in theory this would be the one time that this error message gets printed (other threads then calling `.get_or_initialize()` would early-return an `Err` above)

And in practice (eg I added a test for this) I've only been seeing it get printed once

However I think it's actually "race"-y because I think another thread could enter this block before this one has had a chance to `.mark_failed()` here

Ooh actually I bet I could use a "fetch-swap" or whatever in `.mark_failed()` so basically ensure that after the call to `.mark_failed()` it's definitely stored the failure value but then return a value from that call to `.mark_failed()` that indicates whether we actually "won" the race to insert that failure value. And then only print the failure message here if we won

src/lib.rs
```diff
@@ -91,25 +99,77 @@ fn get_output_mode(args: &Args) -> OutputMode {
...
-    let supported_language = args.language.get_language();
-    let language = supported_language.language();
-    let query = Arc::new(get_query(&query_source, language));
-    let capture_index = args.capture_name.as_ref().map_or(0, |capture_name| {
-        query
-            .capture_index_for_name(capture_name)
-            .expect(&format!("Unknown capture name: `{}`", capture_name))
-    });
+    let specified_supported_language = args.language.map(|language| language.get_language());
+    let query_or_failure_by_language: Mutex<HashMap<SupportedLanguageName, Option<Arc<Query>>>> =
```
helixbass, line 154: A `None` value in this `HashMap` indicates "we already tried parsing the query with this language's grammar and it failed to parse"

src/lib.rs
```diff
@@ -91,25 +99,77 @@ fn get_output_mode(args: &Args) -> OutputMode {
...
     let output_mode = get_output_mode(&args);
     let buffer_writer = BufferWriter::stdout(ColorChoice::Never);
 
-    get_project_file_walker(&*supported_language, &args.use_paths())
+    get_project_file_walker(specified_supported_language.as_deref(), &args.use_paths())
         .into_parallel_iterator()
         .for_each(|project_file_dir_entry| {
+            let language = maybe_supported_language_from_path(project_file_dir_entry.path())
+                .expect("Walker should've been pre-filtered to just supported file types");
+            let query = return_if_none!(get_and_cache_query_for_language(
```
helixbass, line 165: Reading this I like this macro better than the boilerplate of
```
let whatever = ...;
if whatever.is_none() {
    return;
}
let whatever = whatever.unwrap();
```
or:
```
let whatever = match ... {
    None => return,
    Some(whatever) => whatever,
};
```
but it would be nice in terms of "noisiness"/"precedence" to write it postfix eg:
```
let query = ...
    .return_if_none!();
```
Looks like that is a [proposal](https://github.com/rust-lang/rfcs/pull/2442)

Or I guess if this function returned an `Option`/`Result` then could use `?`?

src/lib.rs
```diff
@@ -129,6 +189,38 @@ pub fn run(args: Args) {
                 .unwrap();
             buffer_writer.print(printer.get_mut()).unwrap();
         });
+
+    error_if_no_successful_query_parsing(&query_or_failure_by_language);
```
helixbass, line 193: Turn "n lazy soft fails" into a "hard fail"

src/lib.rs
```diff
@@ -91,25 +99,77 @@ fn get_output_mode(args: &Args) -> OutputMode {
...
 
-    get_project_file_walker(&*supported_language, &args.use_paths())
+    get_project_file_walker(specified_supported_language.as_deref(), &args.use_paths())
         .into_parallel_iterator()
         .for_each(|project_file_dir_entry| {
+            let language = maybe_supported_language_from_path(project_file_dir_entry.path())
+                .expect("Walker should've been pre-filtered to just supported file types");
+            let query = return_if_none!(get_and_cache_query_for_language(
+                &query_source,
+                &query_or_failure_by_language,
```
helixbass, line 167: There is currently no optimization for when you set the `--language` flag other than the fact that the walker limits itself to just returning files of that type

ie we're still using this `Mutex<HashMap<...>>` even when we "know" that there's only one language/parsed query we're interested in

On the one hand it seemed a little silly to tear out that "fast path" but it's hard to picture how it wouldn't have made the code more complex and so I deemed it premature optimization to try and maintain that

src/lib.rs
```diff
@@ -129,6 +189,38 @@ pub fn run(args: Args) {
...
+fn get_and_cache_query_for_language(
+    query_source: &str,
+    query_or_failure_by_language: &Mutex<HashMap<SupportedLanguageName, Option<Arc<Query>>>>,
+    language: &dyn SupportedLanguage,
+) -> Option<Arc<Query>> {
+    query_or_failure_by_language
+        .lock()
+        .unwrap()
+        .entry(language.name())
+        .or_insert_with(|| maybe_get_query(query_source, language.language()).map(Arc::new))
```
helixbass, line 222: I thought a little bit about this in terms of performance - I don't know how slow query-parsing is (vs project file iteration) but there could be competition for this mutex especially "at the beginning" (if a bunch of project files get into `rayon`-world and get stuck waiting for the mutex to get access to the parsed query for their language)

But (in the "effectively single-language" scenario) they'd all have to in some sense wait for that query-parsing time anyway so doesn't seem terrible

If there are multiple supported languages being encountered then you could have parsing less-than-optimally parallelized I think, seems like then maybe having per-language mutexes would be the optimization there (rather than one mutex for n supported languages)?

tests/output.rs
```diff
@@ -86,6 +86,25 @@ fn do_sorted_lines_match(actual_output: &str, expected_output: &str) -> bool {
...
+        output,
+    } = parse_command_and_output(command_and_output);
+    let command_name = command_line_args.remove(0);
+    Command::cargo_bin(&command_name)
+        .unwrap()
+        .args(command_line_args)
+        .current_dir(get_fixture_dir_path_from_name(fixture_dir_name))
+        .assert()
+        .failure()
+        // .stderr(predicate::function(|stderr: &str| {
```
helixbass, line 101: Since it's not always super easy to visually parse out the expected vs actual when this fails (let alone how they're different) it's tempting to leave this "debug helper" commented-out here like this for quick toggling

Really it would be nice if it showed a "pretty" diff of expected vs actual

tests/output.rs
```diff
@@ -227,3 +246,94 @@ fn test_specify_multiple_files() {
...
+}
+
+#[test]
+fn test_auto_language_multiple_parseable_languages() {
+    assert_sorted_output(
+        "mixed_project",
+        r#"
+            $ tree-sitter-grep --query-source '(arrow_function) @arrow_function'
+            javascript_src/index.js:1:const js_foo = () => {}
+            typescript_src/index.tsx:1:const foo = () => {}
```
helixbass, line 325: Pretty cool cross-language queries!

(no idea how often that would be practically useful but sort of technically interesting)

peterstuart (approved):
I'm not sure how it'd affect performance (probably faster in some ways, slower in others), but I think it'd be simpler to try and parse the queries against all supported languages before starting the path walking, so:

- don't need to iterate over files that aren't needed
- don't have to deal with lazily parsing the queries per thread
- could print some info before the results about which languages it's trying to use (maybe on stderr so it doesn't "dirty" up the stdout results)

WDYT?

src/lib.rs
helixbass: Ok I updated this to use `.compare_exchange()` to avoid such a race. And in fact "validated" it by artificially introducing a long sleep, seeing the race happen (ie multiple printings of the failure message), and then no longer seeing it happen after adding the `.compare_exchange()`

src/lib.rs
```diff
@@ -83,25 +94,101 @@ fn get_output_mode(args: &Args) -> OutputMode {
     }
 }
 
+struct MaybeInitializedCaptureIndex(AtomicU32);
+
+impl MaybeInitializedCaptureIndex {
+    const UNINITIALIZED: u32 = u32::MAX;
+
+    const FAILED: u32 = u32::MAX - 1;
```
helixbass, line 102: These `u32::MAX`/`u32::MAX - 1`'s (while encapsulated) were seeming a bit "strewn about" so added named constants for them

src/lib.rs
```diff
@@ -83,25 +94,101 @@ fn get_output_mode(args: &Args) -> OutputMode {
...
+            Some(capture_name) => {
+                let capture_index = query.capture_index_for_name(capture_name);
+                if capture_index.is_none() {
+                    let did_mark_failed = self.mark_failed();
+                    if did_mark_failed {
+                        fail(&format!("invalid capture name '{}'", capture_name));
+                    } else {
+                        // whichever other thread "won the race" will have called this fail()
+                        // so we'll be getting killed shortly?
+                        thread::sleep(Duration::from_millis(100_000));
```
helixbass, line 146: Let me know if you disagree with this, I am certainly not seeing this cause anything to actually visibly sleep for 100s

helixbass (commented):
> I'm not sure how it'd affect performance (probably faster in some ways, slower in others), but I think it'd be simpler to try and parse the queries against all supported languages before starting the path walking, so:
>
> - don't need to iterate over files that aren't needed
> - don't have to deal with lazily parsing the queries per thread
> - could print some info before the results about which languages it's trying to use (maybe on stderr so it doesn't "dirty" up the stdout results)
>
> WDYT?

Ya that is certainly intriguing. I opened an issue. All of this may be "naive about optimization guy" but my instinct tends to go the other way:

I sort of like the lazy parsing

And seems like if the # of supported languages grows (I guess I'm thinking it would 10-20-ish to start? I think I'd said something like "all languages with tree-sitter grammar/parser crates" but that's actually a lot (a lot of which are obscure languages) so I don't know "where's the exact right place to land" as far as which to bundle with such a bundling approach) that would only increase that "up-front cost"

And if the "we're in effectively a single-language project but just didn't want to write out the `--language` flag" case would in fact be common (as I'd guess it would to the point of being "predominant") then it seems like you're just incurring up-front cost of trying to parse the query in n additional languages (which is something even if parallelized) since you wouldn't really be "stripping anything out"/saving anything in terms of iterating over less files (that would only apply in a directory where there are multiple types of supported files but the query only parses for a subset of them?)

> could print some info before the results about which languages it's trying to use (maybe on stderr so it doesn't "dirty" up the stdout results)

This is valid but (even if on stderr) I think could still be just "noise" for a lot of people (like "ya no shit it's a Rust query, I just want to see the matches"), but others would probably like it?

PR #28, opened 2023-06-11
Some failure tests

helixbass:
In this PR:
- add some more tests for eg command-line parsing failures
- a couple changes from code review on previous PR's

To test:
Things should still work, tests should pass

### DIFF
--- a/tests/output.rs
+++ b/tests/output.rs
@@ -21,16 +21,17 @@ fn get_fixture_dir_path_from_name(fixture_dir_name: &str) -> PathBuf {
 
 fn parse_command_and_output(command_and_output: &str) -> CommandAndOutput {
     let mut lines = command_and_output.split("\n").collect::<Vec<_>>();
-    if !lines.is_empty() {
-        if regex!(r#"^\s*$"#).is_match(&lines[0]) {
-            lines.remove(0);
-        }
+    if lines.is_empty() {
+        panic!("Expected at least a command line");
+    }
+    if lines[0].trim().is_empty() {
+        lines.remove(0);
     }
     let command_line = lines.remove(0);
     let indent = regex!(r#"^\s*"#).find(&command_line).unwrap().as_str();
     let command_line_args = parse_command_line(strip_indent(&command_line, indent));
     if !lines.is_empty() {
-        if regex!(r#"^\s*$"#).is_match(&lines[lines.len() - 1]) {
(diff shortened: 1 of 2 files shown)


tests/output.rs
```diff
@@ -23,24 +23,29 @@ fn get_fixture_dir_path_from_name(fixture_dir_name: &str) -> PathBuf {
...
             lines.pop();
         }
     }
     let output: String = lines
         .into_iter()
         .map(|line| {
-            assert!(line.starts_with(indent));
-            format!("{}\n", strip_indent(line, indent))
+            if line.is_empty() {
+                "\n".to_owned()
```
helixbass, line 44: This is just to not have to add `indent` spaces at the beginning of blank output lines (since my editor wasn't inclined to do that)

tests/output.rs
```diff
@@ -339,3 +360,73 @@ fn test_auto_language_single_parseable_languages() {
         "#,
     );
 }
+
+#[test]
+fn test_capture_name() {
+    assert_sorted_output(
+        "rust_project",
+        r#"
+            $ tree-sitter-grep --query-source '(function_item name: (identifier) @name) @function_item' --language rust --capture function_item
```
peterstuart, line 369: Oh, I guess I missed that you can specify the capture as an option. Is the reason for that (rather than just not capturing things you don't care about) that you might use other captures in predicates?

tests/output.rs
```diff
@@ -339,3 +360,73 @@ fn test_auto_language_single_parseable_languages() {
...
+        "#,
+    );
+}
+
+#[test]
+fn test_predicate() {
+    assert_sorted_output(
+        "rust_project",
+        r#"
+            $ tree-sitter-grep --query-source '(function_item name: (identifier) @name (#eq? @name "add")) @function_item' --language rust --capture function_item
```
peterstuart, line 388: Seems like the answer is "yes" 😄

tests/output.rs
helixbass: Ya I guess that's probably the most common reason you'd have multiple captures? Or maybe if you were eg using a "pre-written" query?

Related-ish, I'm pretty sure that it assigns "capture indices" based on the order in which the capture names actually appear in the query (so the default would always be the capture that appears first), so here this is in fact testing something non-default (I think I ran this without the `--capture` argument to sanity-check that and did in fact not see the multiple lines corresponding to entire functions being printed out)

PR #29, opened 2023-06-12
Switch to the standard library OnceCell/OnceLock

### LINKED ISSUE #25: Use new std version of OnceCell?
Per [this comment](https://github.com/helixbass/tree-sitter-grep/pull/20/files#r1225782001), sounds like we could use the [newly stabilized](https://blog.rust-lang.org/2023/06/01/Rust-1.70.0.html#oncecell-and-oncelock) standard-library version of `OnceCell` in at least some places?

peterstuart:
Close #25.


helixbass (approved):
@peterstuart lgtm

PR #30, opened 2023-06-12
Run GitHub Actions on Ubuntu and Windows

peterstuart:
@helixbass It looks to me like everything basically works (in that the output I see in the tests look reasonable), but the test code is going to have to be able standardize the output before comparing it against the snapshots. That looks like it might just be:
- stripping out carriage returns ("\r") at the end of lines
- handling back- vs forward-slashes

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,6 +21,8 @@ jobs:
           rustup toolchain install nightly --allow-downgrade -c rustfmt
       - uses: Swatinem/rust-cache@v2
       - run: cargo +nightly fmt -- --check
+        if: matrix.os == 'ubuntu-latest'
       - run: cargo clippy --all-targets --tests -- -D warnings
       - run: cargo test
       - run: RUSTDOCFLAGS='--deny warnings' cargo doc --no-deps
+        if: matrix.os == 'ubuntu-latest'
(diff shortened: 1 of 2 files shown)


helixbass (approved):
@peterstuart thanks for setting this up, I pushed a couple commits getting tests passing on CI (and locally at least in a Mac Parallels Desktop Windows setup), looks like one of the other jobs is failing on Windows: https://github.com/helixbass/tree-sitter-grep/actions/runs/5245146544/jobs/9472179086?pr=30

Does that even need to get to run on Windows for CI? Or is running the tests the only relevant thing to check on Windows? Also Windows CI appeared to be running slower so omitting eg `cargo fmt`/`cargo clippy` on Windows might speed it up. Although I guess Clippy could spit out different warnings on Windows since there is different code active so maybe it should legitimately get run in CI?

peterstuart:
> Does that even need to get to run on Windows for CI? Or is running the tests the only relevant thing to check on Windows? Also Windows CI appeared to be running slower so omitting eg cargo fmt/cargo clippy on Windows might speed it up. Although I guess Clippy could spit out different warnings on Windows since there is different code active so maybe it should legitimately get run in CI?

Yeah, I think it's worth running clippy still, fmt and docs seem pointless. I can update the workflow. I think most of the reason that it's running slower on Windows at the moment is that it doesn't cache all the built dependencies on failed builds, and we haven't had a successful one yet, so it's doing a clean build on every attempt.

PR #32, opened 2023-06-12
Set up Dependabot

### LINKED ISSUE #31: Set up Dependabot?
I've never done this but I think it's pretty declarative eg add a `dependabot.yml`?

peterstuart:
Close #31.

I don't think you can see this working without merging it, but I copied it from another working project, so should be fine?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 1000


.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 1000
```
peterstuart, line 7: Without this it defaults to some very small limit. I'm not sure why you wouldn't just want to see all at once?

helixbass: Ya that is a weird default. I guess it can get "noisy" but whatever that does not seem like the proper way to mitigate that

helixbass (approved):
@peterstuart lgtm thanks for setting it up

PR #34, opened 2023-06-13
Test filter plugins

helixbass:
In this PR:
- add some "success case" tests for filter plugins packaged as `examples/`

Not in this PR:
- it seems worth trying to make the "plugin failed because of missing/invalid filter argument" case fail more nicely (vs panicking in FFI-world), so maybe that involves eg encoding a failure code in the return value of the `initialize()` plugin function and translating that into a `Result`? But that sounds like a bit of a refactor so kept it separate

To test:
You should be able to build the examples in `examples` as normally expected (eg `cargo build --example whatever`)
If you completely remove your local `target/` directory and then run `cargo test` it should pass (implying that it successfully built the filter plugins while running `cargo test`, I think)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,5 +27,14 @@ name = "tree-sitter-grep"
 
 [dev-dependencies]
 assert_cmd = "2.0.11"
+escargot = "0.5.7"
 predicates = "3.0.3"
 shlex = "1.1.0"
+
+[[example]]
+name = "filter_before_line_number"
+crate-type = ["cdylib"]
+
+[[example]]
+name = "filter_before_line_10"
+crate-type = ["cdylib"]
--- a/examples/filter_before_line_number.rs
+++ b/examples/filter_before_line_number.rs
@@ -1,0 +1,25 @@
+use std::{
+    ffi::CStr,
+    sync::atomic::{AtomicUsize, Ordering},
+};
+
+use libc::c_char;
+use tree_sitter::Node;
+
+static ROW_NUMBER: AtomicUsize = AtomicUsize::new(0);
+
+#[no_mangle]
+pub extern "C" fn initialize(value: *const c_char) {
+    assert!(!value.is_null(), "Expected filter argument");
+    let value: usize = unsafe { CStr::from_ptr(value) }
+        .to_str()
+        .unwrap()
+        .parse()
+        .expect("Expected filter argument to be a usize");
+    ROW_NUMBER.store(value, Ordering::Relaxed);
+}
+
+#[no_mangle]
+pub extern "C" fn filterer(node: &Node) -> bool {
+    node.start_position().row < ROW_NUMBER.load(Ordering::Relaxed)
+}
--- a/examples/filter_before_line_number.rs
+++ b/examples/filter_before_line_number.rs
@@ -8,6 +8,7 @@ use tree_sitter::Node;
 
 static ROW_NUMBER: AtomicUsize = AtomicUsize::new(0);
 
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
 #[no_mangle]
 pub extern "C" fn initialize(value: *const c_char) {
     assert!(!value.is_null(), "Expected filter argument");
--- a/tests/output.rs
+++ b/tests/output.rs
@@ -67,9 +67,24 @@ fn strip_indent<'line>(line: &'line str, indent: &str) -> &'line str {
     &line[indent.len()..]
 }
 
+#[cfg(target_os = "macos")]
+const DYNAMIC_LIBRARY_EXTENSION: &str = ".dylib";
+#[cfg(all(unix, not(target_os = "macos")))]
+const DYNAMIC_LIBRARY_EXTENSION: &str = ".so";
+#[cfg(windows)]
+const DYNAMIC_LIBRARY_EXTENSION: &str = ".dll";
+
 fn parse_command_line(command_line: &str) -> Vec<String> {
     assert!(command_line.starts_with('$'));
-    shlex::split(&command_line[1..]).unwrap()
+    shlex::split(&command_line[1..])
+        .unwrap()
+        .iter()
+        .map(|arg| {
+            regex!(r#"\.so$"#)
+                .replace(arg, DYNAMIC_LIBRARY_EXTENSION)
+                .into_owned()
+        })
+        .collect()
 }
 
 fn assert_sorted_output(fixture_dir_name: &str, command_and_output: &str) {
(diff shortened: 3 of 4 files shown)


Cargo.toml
```diff
@@ -27,5 +27,14 @@ name = "tree-sitter-grep"
 
 [dev-dependencies]
 assert_cmd = "2.0.11"
+escargot = "0.5.7"
 predicates = "3.0.3"
 shlex = "1.1.0"
+
+[[example]]
+name = "filter_before_line_number"
+crate-type = ["cdylib"]
```
helixbass, line 36: Per the [docs](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples), specify that these examples are shared libraries

examples/filter_before_line_number.rs
```diff
@@ -0,0 +1,26 @@
...
+    ffi::CStr,
+    sync::atomic::{AtomicUsize, Ordering},
+};
+
+use libc::c_char;
+use tree_sitter::Node;
+
+static ROW_NUMBER: AtomicUsize = AtomicUsize::new(0);
+
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
```
helixbass, line 11: Clippy was arguing that this whole `fn` should be marked `unsafe` because of the `CStr::from_ptr()` dereferencing of a raw pointer inside it

But I don't understand well enough what the argument/implications are (this is only being consumed via FFI which is `unsafe`-aware, no?) so didn't want to blindly go with it

tests/output.rs
```diff
@@ -67,9 +67,24 @@ fn strip_indent<'line>(line: &'line str, indent: &str) -> &'line str {
     &line[indent.len()..]
 }
 
+#[cfg(target_os = "macos")]
+const DYNAMIC_LIBRARY_EXTENSION: &str = ".dylib";
+#[cfg(all(unix, not(target_os = "macos")))]
```
helixbass, line 72: These sort of seemed like the correct way to specify these? Let's see if CI passes for Linux/Windows (I was testing locally on Mac and that seemed to be working)

tests/output.rs
```diff
@@ -150,6 +165,14 @@ fn massage_error_output(output: &str) -> String {
     output.replace(".exe", "")
 }
 
+fn build_example(example_name: &str) {
+    // CargoBuild::new().example(example_name).exec().unwrap();
```
helixbass, line 169: This was weird, per [this answer](https://stackoverflow.com/a/75564431/732366) I tried using [`escargot`](https://docs.rs/escargot/latest/escargot/index.html) to "properly" tell it to build a certain example but this seemed to be hanging when I ran the tests

So fell back to just manually creating a `cargo` command line? The one thing I notice there is that you see some "Blocking waiting for file lock on build directory"'s in the `cargo test` output but doesn't seem to actively cause any problems

Cargo.toml
peterstuart: Interesting, I wouldn't have guessed that a dynamic library was a valid type of "example".

examples/filter_before_line_number.rs
peterstuart: I guess any function accepting C-pointers (and doing something with them) is kind of unsafe by definition, but on the other hand, who cares if it's essentially the entry point to the whole library. I don't really see a downside to marking it unsafe though. 🤷🏻

tests/output.rs
peterstuart: Ah interesting, so I guess BSD, etc fall under this too (not just Linux)?

tests/output.rs
```diff
@@ -67,9 +66,35 @@ fn strip_indent<'line>(line: &'line str, indent: &str) -> &'line str {
     &line[indent.len()..]
 }
 
+#[cfg(target_os = "macos")]
+const DYNAMIC_LIBRARY_EXTENSION: &str = ".dylib";
+#[cfg(all(unix, not(target_os = "macos")))]
+const DYNAMIC_LIBRARY_EXTENSION: &str = ".so";
+#[cfg(windows)]
+const DYNAMIC_LIBRARY_EXTENSION: &str = ".dll";
```
peterstuart, line 74: You could make this slightly less repetitive by using the `cfg!` macro (as long as if-expressions can be `const`, don't know if that's true): https://doc.rust-lang.org/std/macro.cfg.html

tests/output.rs
```diff
@@ -67,9 +66,35 @@ fn strip_indent<'line>(line: &'line str, indent: &str) -> &'line str {
...
+
+#[cfg(unix)]
+fn get_dynamic_library_name(library_name: &str) -> String {
+    format!("lib{library_name}{}", DYNAMIC_LIBRARY_EXTENSION)
+}
+#[cfg(windows)]
+fn get_dynamic_library_name(library_name: &str) -> String {
+    format!("{library_name}{}", DYNAMIC_LIBRARY_EXTENSION)
+}
```
peterstuart, line 83: And same here, I think it'd be a little cleaner to write an `if cfg!` inside the function rather than repeating the function definition.

tests/output.rs
```diff
@@ -67,9 +66,35 @@ fn strip_indent<'line>(line: &'line str, indent: &str) -> &'line str {
...
+
 fn parse_command_line(command_line: &str) -> Vec<String> {
     assert!(command_line.starts_with('$'));
-    shlex::split(&command_line[1..]).unwrap()
+    shlex::split(&command_line[1..])
+        .unwrap()
+        .iter()
+        .map(|arg| {
+            regex!(r#"lib(\S+)\.so$"#)
```
peterstuart, line 91: This seems like it's relying on the `.so` extension? How does that work?

peterstuart (approved):

Cargo.toml
helixbass: Ya the author of [this article](http://xion.io/post/code/rust-examples.html#fn:1) explicitly says he thinks that isn't useful but _boy was he wrong_

tests/output.rs
helixbass: Right and hopefully they all also use `.so` (at least if they want to run this test suite successfully)

tests/output.rs
helixbass: Ah interesting ok updated to use `cfg!()` for all of these

> as long as if-expressions can be `const`, don't know if that's true

oh ok looking at those docs that actually isn't required because it doesn't `const`-eval the whole `if`-block, just replaces the `cfg!(...)` -> boolean

tests/output.rs
helixbass: Because we control the tests?

I guess how I'm thinking of it is that the tests are written "canonically" (for Linux?) and then we massage from there based on what actual platform it's running on

(I'm also guessing you just had it "backwards" and didn't really need that explanation)

PR #35, opened 2023-06-13
Test filter plugin: failure

helixbass:
In this PR:
- make the `initialize()` filter plugin function return a value that encodes whether/how it succeeded/failed
- tests for that

To test:
You should be able to manually mimic what the newly added tests are doing as far as returning an error from a filter plugin if it isn't given an argument that it expects that gets printed "cleanly"

Based on `test-filter-plugin`

### DIFF
--- a/examples/filter_before_line_number.rs
+++ b/examples/filter_before_line_number.rs
@@ -5,10 +5,7 @@ use std::{
 
 use libc::c_char;
 use tree_sitter::Node;
-use tree_sitter_grep::{
-    PluginInitializeReturn, PLUGIN_INITIALIZE_ARGUMENT_NOT_PARSEABLE,
-    PLUGIN_INITIALIZE_MISSING_EXPECTED_ARGUMENT, PLUGIN_INITIALIZE_SUCCEEDED,
-};
+use tree_sitter_grep::PluginInitializeReturn;
 
 static ROW_NUMBER: AtomicUsize = AtomicUsize::new(0);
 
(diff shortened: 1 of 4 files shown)


examples/filter_before_line_number.rs
```diff
@@ -5,19 +5,27 @@ use std::{
 
 use libc::c_char;
 use tree_sitter::Node;
+use tree_sitter_grep::{
+    PluginInitializeReturn, PLUGIN_INITIALIZE_ARGUMENT_NOT_PARSEABLE,
+    PLUGIN_INITIALIZE_MISSING_EXPECTED_ARGUMENT, PLUGIN_INITIALIZE_SUCCEEDED,
```
helixbass, line 10: So this obviously wants to be an enum but I didn't know how/if you could use a Rust enum as a C-FFI type? So went for the moment with "how maybe you're supposed do it" via a simple type alias for `u32` + named constants for the different expected error/success codes?

peterstuart (approved):

examples/filter_before_line_number.rs
helixbass: Ok updated to use a `#[repr(u8)]` enum instead

So if someone wanted to write a plugin in something other than Rust then they wouldn't be able to import the Rust types anyway (so that didn't make sense really)?

So they would just follow documentation that said "return 0 for success, 1 for missing argument and 2 for unparseable argument (where the returned type is the C equivalent of `u8` (`unsigned char`?))"?

src/lib.rs
```diff
@@ -37,6 +37,7 @@ use language::{
     SupportedLanguageName,
 };
 use plugin::get_loaded_filter;
+pub use plugin::PluginInitializeReturn;
```
helixbass, line 40: So reading this this is now "surface area for `tree_sitter_grep`-the-crate"?

I'm remembering that in some of those FFI-plugin-examples they were extracting this type of thing out into a shared "plugin" crate, maybe that is a good idea (I'll add an issue)?

examples/filter_before_line_number.rs
peterstuart: Yeah, makes sense

PR #38, opened 2023-06-14
Support "no query" (with filter) mode

helixbass:
In this PR:
- allow not passing any "query" argument if you pass a filter argument (and use a default "pass all AST nodes through to the filter" query in that case)
- futz with `clap` a little bit

To test:
You should be able to build + run example filters (I added a new one which at least limits the AST node types that it produces matches for, I didn't try running in non-query mode for one of the existing ones but that is presumably a lot of AST nodes matched) with or without also passing a query argument
Per the tests the output produced by `clap` eg when doing `--help`, not passing required arguments or passing an unrecognized argument shouldn't be too insane/unhelpful

### DIFF
--- a/examples/filter_function_items_before_line_10.rs
+++ b/examples/filter_function_items_before_line_10.rs
@@ -1,0 +1,6 @@
+use tree_sitter::Node;
+
+#[no_mangle]
+pub extern "C" fn filterer(node: &Node) -> bool {
+    node.kind() == "function_item" && node.start_position().row < 10
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,7 +16,7 @@ use std::{
     time::Duration,
 };
 
-use clap::Parser;
+use clap::{ArgGroup, Parser};
 use grep::{
     matcher::{Match, Matcher, NoCaptures, NoError},
     printer::{Standard, StandardBuilder},
@@ -41,17 +41,30 @@ pub use plugin::PluginInitializeReturn;
 use treesitter::{get_matches, maybe_get_query};
 
 #[derive(Parser)]
+#[clap(group(
+    ArgGroup::new("query_or_filter")
+        .multiple(true)
+        .required(true)
+        .args(&["path_to_query_file", "query_source", "filter"])
+))]
 pub struct Args {
     pub paths: Vec<PathBuf>,
-    #[command(flatten)]
-    pub query_args: QueryArgs,
+    // #[arg(short = 'Q', long = "query-file", conflicts_with = "query_source",
+    // required_unless_present_any = ["query_source", "filter"])]
+    #[arg(short = 'Q', long = "query-file", conflicts_with = "query_source")]
+    pub path_to_query_file: Option<PathBuf>,
+    // #[arg(short, long, conflicts_with = "path_to_query_file", required_unless_present_any =
+    // ["path_to_query_file", "filter"])]
+    #[arg(short, long, conflicts_with = "path_to_query_file")]
+    pub query_source: Option<String>,
     #[arg(short, long = "capture")]
     pub capture_name: Option<String>,
     #[arg(short, long, value_enum)]
     pub language: Option<SupportedLanguageName>,
+    // #[arg(short, long, required_unless_present_any = ["path_to_query_file", "query_source"])]
     #[arg(short, long)]
     pub filter: Option<String>,
-    #[arg(short = 'a', long)]
+    #[arg(short = 'a', long, requires = "filter")]
     pub filter_arg: Option<String>,
     #[arg(long)]
     pub vimgrep: bool,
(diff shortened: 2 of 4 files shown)


examples/filter_function_items_before_line_10.rs
```diff
@@ -0,0 +1,6 @@
+use tree_sitter::Node;
+
+#[no_mangle]
+pub extern "C" fn filterer(node: &Node) -> bool {
+    node.kind() == "function_item" && node.start_position().row < 10
```
helixbass, line 5: Performance-wise I didn't try anything that spit out an insane # of matches but for something like this that produces a reasonable # of matches I am seeing it run almost as fast vs the equivalent query + filter combo*

Which seems insane, but classic incorrect guess about optimization/performance I guess

[*] if you set the `--language` flag that is - there seems to be some kind of performance issue with the combination of no-query + auto-language. Which also seems to go against my mental model of "the only known file-types it should be walking over is just Rust files which is what happens with passing `--language rust` so should be basically exactly the same"?

src/lib.rs
```diff
@@ -41,17 +41,30 @@ pub use plugin::PluginInitializeReturn;
 use treesitter::{get_matches, maybe_get_query};
 
 #[derive(Parser)]
+#[clap(group(
```
helixbass, line 44: The objective here is to convince `clap` that (a) `--query-source` + `--query-file` are mutually exclusive (b) you can't omit all of `--query-source`, `--query-file` and `--filter`

I tried a couple different ways of expressing that (see eg the commented-out lines below) but while this produces some wonky `clap` "help"-style output (see some tests below) it seemed better than the alternatives I was able to figure out

src/lib.rs
```diff
@@ -41,17 +41,30 @@ pub use plugin::PluginInitializeReturn;
...
+    pub query_source: Option<String>,
     #[arg(short, long = "capture")]
     pub capture_name: Option<String>,
     #[arg(short, long, value_enum)]
     pub language: Option<SupportedLanguageName>,
+    // #[arg(short, long, required_unless_present_any = ["path_to_query_file", "query_source"])]
     #[arg(short, long)]
     pub filter: Option<String>,
-    #[arg(short = 'a', long)]
+    #[arg(short = 'a', long, requires = "filter")]
```
helixbass, line 67: I should/meant to add a test for this (providing `--filter-arg` without `--filter` should be an error)

tests/output.rs
```diff
@@ -159,12 +159,34 @@ fn assert_failure_output(fixture_dir_name: &str, command_and_output: &str) {
         }));
 }
 
+fn assert_non_match_output(fixture_dir_name: &str, command_and_output: &str) {
```
helixbass, line 162: `--help` prints to `stdout` (not `stderr`) but we don't want to do the matches-style "output sorting" so this basically mimics `assert_failure_output()` but looks at `stdout` instead of `stderr`

tests/output.rs
```diff
@@ -159,12 +159,34 @@ fn assert_failure_output(fixture_dir_name: &str, command_and_output: &str) {
...
+}
+
 fn massage_error_output(output: &str) -> String {
     if cfg!(windows) {
         output.replace(".exe", "")
     } else {
         output.to_owned()
     }
+    .split('\n')
+    .map(|line| line.trim_end())
```
helixbass, line 187: There was some weird trailing whitespace being emitted by `clap` in the `--help` message (so strip it here)

tests/output.rs
```diff
@@ -487,7 +509,7 @@ fn test_unknown_option() {
 
               tip: a similar argument exists: '--query-source'
 
-            Usage: tree-sitter-grep <--query-file <PATH_TO_QUERY_FILE>|--query-source <QUERY_SOURCE>> [PATHS]...
+            Usage: tree-sitter-grep <--query-file <PATH_TO_QUERY_FILE>|--query-source <QUERY_SOURCE>|--filter <FILTER>> <PATHS|--query-file <PATH_TO_QUERY_FILE>|--query-source <QUERY_SOURCE>|--capture <CAPTURE_NAME>|--language <LANGUAGE>|--filter <FILTER>|--filter-arg <FILTER_ARG>|--vimgrep>
```
helixbass, line 512: This is the most crazy one with the current `clap` configuration

tests/output.rs
```diff
@@ -550,3 +572,75 @@ fn test_filter_plugin_unparseable_argument() {
...
+    );
+}
+
+#[test]
+fn test_help_option() {
+    assert_non_match_output(
+        "rust_project",
+        r#"
+            $ tree-sitter-grep --help
+            Usage: tree-sitter-grep [OPTIONS] <--query-file <PATH_TO_QUERY_FILE>|--query-source <QUERY_SOURCE>|--filter <FILTER>> [PATHS]...
```
helixbass, line 614: While this could be considered technically slightly misleading (it makes it look like `--filter` is mutually exclusive with the query arguments) it's pretty close and I'm not really sure how you'd even write the actual logic out in a format like this?

examples/filter_function_items_before_line_10.rs
peterstuart: Hmm, yeah I don't really understand that.

peterstuart (approved):
The only thing I can think of that could help with the clap issue is to remove one of the query params, and try to figure out at runtime whether they passed in a path or a query. That seems possible (at least if you don't care if someone names files in a crazy way like `(my_filename)`), but probably kind of confusing for the user?

helixbass (commented):
> The only thing I can think of that could help with the clap issue is to remove one of the query params, and try to figure out at runtime whether they passed in a path or a query. That seems possible (at least if you don't care if someone names files in a crazy way like `(my_filename)`), but probably kind of confusing for the user?

Ya agreed if it was a usability improvement could be an option but I agree it's moreso potentially confusing

PR #40, opened 2023-06-15
Extract modules

helixbass:
In this PR:
- extract a bunch of smaller files from `src/lib.rs`

To test:
Existing behavior should still work

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -26,7 +26,7 @@ pub struct Args {
     #[arg(short = 'a', long, requires = "filter")]
     pub filter_arg: Option<String>,
     #[arg(long)]
-    pub vimgrep: bool,
+    vimgrep: bool,
 }
 
 impl Args {
--- a/src/args.rs
+++ b/src/args.rs
@@ -12,7 +12,7 @@ use crate::language::SupportedLanguageName;
         .args(&["path_to_query_file", "query_source", "filter"])
 ))]
 pub struct Args {
-    pub paths: Vec<PathBuf>,
+    paths: Vec<PathBuf>,
 
     #[arg(short = 'Q', long = "query-file", conflicts_with = "query_source")]
     pub path_to_query_file: Option<PathBuf>,
--- a/src/args.rs
+++ b/src/args.rs
@@ -13,18 +13,25 @@ use crate::language::SupportedLanguageName;
 ))]
 pub struct Args {
     pub paths: Vec<PathBuf>,
+
     #[arg(short = 'Q', long = "query-file", conflicts_with = "query_source")]
     pub path_to_query_file: Option<PathBuf>,
+
     #[arg(short, long, conflicts_with = "path_to_query_file")]
     pub query_source: Option<String>,
+
     #[arg(short, long = "capture")]
     pub capture_name: Option<String>,
+
     #[arg(short, long, value_enum)]
     pub language: Option<SupportedLanguageName>,
+
     #[arg(short, long)]
     pub filter: Option<String>,
+
     #[arg(short = 'a', long, requires = "filter")]
     pub filter_arg: Option<String>,
+
     #[arg(long)]
     vimgrep: bool,
 }
--- a/src/args.rs
+++ b/src/args.rs
@@ -1,0 +1,58 @@
+use std::path::{Path, PathBuf};
+
+use clap::{ArgGroup, Parser};
+
+use crate::language::SupportedLanguageName;
+
+#[derive(Parser)]
+#[clap(group(
+    ArgGroup::new("query_or_filter")
+        .multiple(true)
+        .required(true)
+        .args(&["path_to_query_file", "query_source", "filter"])
+))]
+pub struct Args {
+    pub paths: Vec<PathBuf>,
+    #[arg(short = 'Q', long = "query-file", conflicts_with = "query_source")]
+    pub path_to_query_file: Option<PathBuf>,
+    #[arg(short, long, conflicts_with = "path_to_query_file")]
+    pub query_source: Option<String>,
+    #[arg(short, long = "capture")]
+    pub capture_name: Option<String>,
+    #[arg(short, long, value_enum)]
+    pub language: Option<SupportedLanguageName>,
+    #[arg(short, long)]
+    pub filter: Option<String>,
+    #[arg(short = 'a', long, requires = "filter")]
+    pub filter_arg: Option<String>,
+    #[arg(long)]
+    pub vimgrep: bool,
+}
+
+impl Args {
+    pub(crate) fn use_paths(&self) -> Vec<PathBuf> {
+        if self.paths.is_empty() {
+            vec![Path::new("./").to_owned()]
+        } else {
+            self.paths.clone()
+        }
+    }
+
+    pub(crate) fn is_using_default_paths(&self) -> bool {
+        self.paths.is_empty()
+    }
+
+    pub(crate) fn output_mode(&self) -> OutputMode {
+        if self.vimgrep {
+            OutputMode::Vimgrep
+        } else {
+            OutputMode::Normal
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub(crate) enum OutputMode {
+    Normal,
+    Vimgrep,
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,102 +1,39 @@
 use std::{
-    cell::{OnceCell, RefCell},
     collections::HashMap,
     fs,
-    path::{Path, PathBuf},
+    path::Path,
     process,
-    rc::Rc,
     sync::{
         atomic::{AtomicU32, Ordering},
-        mpsc,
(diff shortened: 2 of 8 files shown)


src/args.rs
```diff
@@ -0,0 +1,58 @@
...
+        } else {
+            self.paths.clone()
+        }
+    }
+
+    pub(crate) fn is_using_default_paths(&self) -> bool {
+        self.paths.is_empty()
+    }
+
+    pub(crate) fn output_mode(&self) -> OutputMode {
```
helixbass, line 45: Moved this to a method on `Args`

I was seeing some "smart methods" on I believe `ripgrep`'s `Args`-ish type, it seems like a blurry line (of what actually "belongs" on `Args` or not)? But I guess this one felt like a "pure function of the config" so it (+ the `OutputMode` type) maybe seemed appropriate here?

src/lib.rs
```diff
@@ -169,6 +106,30 @@ impl Default for MaybeInitializedCaptureIndex {
 
 const ALL_NODES_QUERY: &str = "(_) @node";
 
+#[derive(Default)]
+struct CachedQueries(Mutex<HashMap<SupportedLanguageName, Option<Arc<Query>>>>);
```
helixbass, line 110: Wanted at least an alias for the long `Mutex<...>` type but then a couple things felt like they could nicely be methods on it so made it a newtype wrapper 

It seems like there are pluses/minuses to "concealing" the fact that this is/contains a `Mutex`?

src/lib.rs
```diff
@@ -178,22 +139,18 @@ pub fn run(args: Args) {
...
-        Default::default();
+    let cached_queries: CachedQueries = Default::default();
     let capture_index = MaybeInitializedCaptureIndex::default();
-    let output_mode = get_output_mode(&args);
+    let output_mode = args.output_mode();
     let buffer_writer = BufferWriter::stdout(ColorChoice::Never);
 
-    get_project_file_walker(specified_supported_language.as_deref(), &args.use_paths())
-        .into_parallel_iterator()
+    get_project_file_parallel_iterator(specified_supported_language.as_deref(), &args.use_paths())
```
helixbass, line 147: As long as this was going in a separate module it started to seem strange to even expose it as a `WalkParallel` (and then explicitly call `.into_parallel_iterator()` on that) (since from this consumer's perspective it can be oblivious to the fact that it ever was a `WalkParallel`) so encapsulated that

src/plugin.rs
```diff
@@ -14,7 +14,7 @@ type RawSymbol<TExportedSymbol> = libloading::os::unix::Symbol<TExportedSymbol>;
 #[cfg(windows)]
 type RawSymbol<TExportedSymbol> = libloading::os::windows::Symbol<TExportedSymbol>;
 
-pub struct Filterer {
+pub(crate) struct Filterer {
```
helixbass, line 17: I guess I don't totally understand the visibility rules, I think there was something where even though the module was imported as non-public in `src/lib.rs` (ie `mod whatever;` not `pub mod whatever;`) it complained that a function inside that module that was marked `pub` was taking/returning a `pub(crate)` type?

So I just made everything `pub(crate)` unless it's "truly" `pub` (ie visible to a consumer of this crate)

src/lib.rs
peterstuart: I'll look at the rest of this PR more closely tomorrow, but this line jumped out at me as something that could be replaced either by a concurrent hashmap (eg https://crates.io/crates/dashmap) or by a `RwLock` around the hashmap. It maybe is not worth pulling in a dependency for this to get a real concurrent hashmap, but even a `RwLock` would probably be an improvement (it only locks when writing, so you can have unlimited concurrent readers).

src/lib.rs
```diff
@@ -169,6 +106,30 @@ impl Default for MaybeInitializedCaptureIndex {
...
+        query_source: &str,
+        language: &dyn SupportedLanguage,
+    ) -> Option<Arc<Query>> {
+        self.0
+            .lock()
+            .unwrap()
+            .entry(language.name())
+            .or_insert_with(|| maybe_get_query(query_source, language.language()).map(Arc::new))
+            .clone()
+    }
```
peterstuart, line 124: This seems like a pretty reasonable use of interior mutability. Since the thing that requires the lock is predictable (it's never going to block for seconds or something), it seems fine to completely hide that from the consumer to me.

peterstuart (approved):

src/args.rs
helixbass: Ah I see `Args` can be an "abstraction" in that sense, sure updated to make `vimgrep` private

src/lib.rs
helixbass: Ok I think I'd commented in the first PR where I added this data structure that "really there should probably be a `Mutex` per-language to be more optimal", so I guess along those lines something "clicked" - we don't actually need a `Mutex`, since (for each language) this is a "lazy-initialize and then just read" pattern, aka (since this is being accessed across multiple threads) a `OnceLock`

So updated to use that and it did indeed compile/work, in order to do that (or similarly the per-language `Mutex` that I'd originally mentioned) you do need to actually initialize all of the potential `HashMap` language "entries" at creation time (or else you'd still need shared (cross-thread) mutable access to the `HashMap` itself and would be back to needing a top-level `Mutex`)

(or to test my understanding say that we didn't want to initialize them all or didn't know all of the potential keys, that's where something like a concurrent hashmap would be an alternative to/avoid needing a top-level `Mutex`?)

src/args.rs
```diff
@@ -0,0 +1,58 @@
...
+    #[arg(short, long = "capture")]
+    pub capture_name: Option<String>,
+    #[arg(short, long, value_enum)]
+    pub language: Option<SupportedLanguageName>,
+    #[arg(short, long)]
+    pub filter: Option<String>,
+    #[arg(short = 'a', long, requires = "filter")]
+    pub filter_arg: Option<String>,
+    #[arg(long)]
+    pub vimgrep: bool,
```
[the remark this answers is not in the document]
helixbass: Sure that is definitely "less dense", updated

src/args.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+impl Args {
+    pub(crate) fn use_paths(&self) -> Vec<PathBuf> {
```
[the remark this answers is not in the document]
helixbass: Sure updated to make `paths` private

I guess I'm trying to distinguish/de-conflate "the paths that were actually supplied on the command-line" vs "internally what list of paths should we actually "use""

For me that somehow seems like a helpful distinction as far as "overall clarity" but I could definitely see the argument (especially with the `paths` field now private) that it should just be named `paths()`

src/lib.rs
helixbass: Ya (per the other comment re updating to use a per-language `OnceLock`) this now seems even more (per I believe a previous comment that "this really shouldn't be that non-optimal because you shouldn't ever have to wait much longer than how long it takes to perform the actual query which you could have to do regardless") like you won't have to wait longer than that since I think the behavior of `OnceLock` is that once it's been initialized then calls to `.get_or_init()` won't block?

src/lib.rs
```diff
@@ -169,6 +109,43 @@ impl Default for MaybeInitializedCaptureIndex {
...
+            .unwrap()
+            .get_or_init(|| maybe_get_query(query_source, language.language()).map(Arc::new))
+            .clone()
+    }
+
+    fn error_if_no_successful_query_parsing(&self) {
+        if !self
+            .0
+            .values()
+            .any(|query| query.get().and_then(|option| option.as_ref()).is_some())
```
helixbass, line 131: In case it is interesting:

I first tried to write this as `query.get().flatten().is_some()`

But that wasn't allowed because technically `.get()` gives back an `Option<&Option<...>>` not an `Option<Option<...>>`

So then I thought "ok either I can `.clone()` the inner `Option` (so `.cloned()` at the top level), but `Arc`-cloning isn't completely "free" (so probably being "naive about optimization guy"), or `.as_ref()` the inner `Option` to go `Option<&Option<...>>` -> `Option<Option<&...>>`" (and then be able to use `.flatten()`)

So I did the latter and then Clippy suggested that `.map(|option| option.as_ref()).flatten()` could be written as `.and_then(|option| option.as_ref())` instead

(conceptually the outer `Option` here is "was the `OnceLock` for this language ever initialized" and the inner `Option` is "given that it was in fact initialized, did the query fail to parse for that language")

src/lib.rs
peterstuart: Yep, this seems perfect.

peterstuart (approved):

src/lib.rs
peterstuart: > (or to test my understanding say that we didn't want to initialize them all or didn't know all of the potential keys, that's where something like a concurrent hashmap would be an alternative to/avoid needing a top-level Mutex?)

Yeah, I think so, and you'd probably still keep the inner `OnceLock`s inside that concurrent hashmap so you're really only locking when necessary?

PR #42, opened 2023-06-18
Use OnceLock for capture index

helixbass:
In this PR:
- get rid of the manual atomics stuff in favor of using a `OnceLock` for the "lazily initialized capture index"

To test:
Existing behavior should still work
Per earlier PR if you want to more actively try and see that the capture index never "double-prints" an error then you can eg artificially add a long delay in the calculation of the capture index (ie `.get_or_init()`) and still see that passing an unknown capture name results in a single error message being printed

Based on `extract-modules`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,10 +3,7 @@ use std::{
     fs,
     path::Path,
     process,
-    sync::{
-        atomic::{AtomicU32, Ordering},
-        Arc, OnceLock,
-    },
+    sync::{Arc, OnceLock},
     thread,
     time::Duration,
 };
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -38,72 +35,29 @@ use project_file_walker::get_project_file_parallel_iterator;
...
-                Ordering::Relaxed,
-                Ordering::Relaxed,
-            );
-            if did_store.is_ok() {
-                return true;
-            }
-        }
-    }
+#[derive(Default)]
+struct CaptureIndex(OnceLock<Result<u32, ()>>);
```
helixbass, line 39: So basically the `CaptureIndex` internally stores a `OnceLock<Result<...>>` to "clue itself in" to the fact that calculating the capture index has already failed, but then since in that case we want to print an error message and immediately kill the process, the "public API" of `CaptureIndex` doesn't actually need to acknowledge the existence of a failure case (so `.get_or_init()` returns plain old `u32`)

It strikes me that if we end up supporting usage as a library/crate (vs just as a command-line executable) then presumably we'd want to return a `Result`/`Err` when they pass something invalid (instead of killing the process)? Which would then mean restructuring some of this code I guess?

src/lib.rs
```diff
@@ -38,72 +35,29 @@ use project_file_walker::get_project_file_parallel_iterator;
...
-                    if did_mark_failed {
-                        fail(&format!("invalid capture name '{}'", capture_name));
-                    } else {
-                        // whichever other thread "won the race" will have called this fail()
-                        // so we'll be getting killed shortly?
-                        thread::sleep(Duration::from_millis(100_000));
-                    }
+impl CaptureIndex {
+    pub fn get_or_init(&self, query: &Query, capture_name: Option<&str>) -> u32 {
+        let mut did_mark_failed = false;
```
helixbass, line 43: My understanding would be that if instead of "hoisting" this state like this we just directly called `fail()` in the failure case (in the inner `.get_or_init()` below) then the `OnceLock` would never actually get initialized to an `Err` (I don't know what actually happens when you call `process::exit()` from inside its `.get_or_init()` callback?) so then other "simultaneous-ish" callers wouldn't nicely see an `Err` and say "ok I'll go to sleep until I get killed"?

src/lib.rs
peterstuart: Ah yeah, that's a good point, but seems fine for now.

src/lib.rs
peterstuart: Good question, I'm not even really sure in general how `process::exit` works in general if you're calling it from something other than the "main" thread of the app. Somehow it coordinates terminating all the threads, or maybe the OS just kills them?

It looks like a regular panic inside `get_or_init` propagates to the caller, but doesn't "ruin" the lock: https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html#panics

(Which is unlike mutexes, I think. Those can be "poisoned" if a panic happens in some cases, and then they can't be used any more.)

peterstuart (approved):
Can't beat deleting a bunch of code to use a nice built-in abstraction instead 🥳

src/lib.rs
peterstuart: https://doc.rust-lang.org/stable/std/sync/struct.PoisonError.html

PR #43, opened 2023-06-21
Sort matches

helixbass:
In this PR:
- fix issue where matches (specifically the captures we specified I guess) were not always in fact returned in sorted order by the underlying tree-sitter query API

To test:
If you run the new test without the corresponding changes for explicitly sorting matches then you should see the test fail (it doesn't include all expected results because we violated our own invariant that our matches are stored in source-text order)


tests/fixtures/sorting_maybe_nesting_related/foo.rs
```diff
@@ -0,0 +1,72 @@
+impl TransformTypeScript {
```
helixbass, line 1: I tried making a little "dummy" test case that I thought had the relevant quality (of the outermost matched AST node being in the reverse order wrt the order of the AST nodes for the capture we specified) but it didn't in fact fail so maybe my understanding of in what cases we were getting "out-of-order" matches wasn't correct

But this is a "real" example that was actually failing (and reproducibly fails here if you don't explicitly sort the matches)

helixbass:
Superceded by #46

PR #45, opened 2023-06-22
Macro contents

helixbass:
In this PR:
- allow querying against the contents of macro invocations (eg `println!(...)` or `vec![...]`)

To test:
If you run a query for eg `(call_expression) @whatever` against a file that contains `vec![call_something()]` it should match the `call_something()`

Based on `sort-matches`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ regex = "1.8.2"
 termcolor = "1.2.0"
 tree-sitter = "0.20.10"
 tree-sitter-javascript = "0.20.0"
-tree-sitter-rust = "0.20.3"
+tree-sitter-rust = { git = "https://github.com/helixbass/tree-sitter-rust", rev = "6327788" }
 tree-sitter-typescript = "0.20.2"
 
 [lib]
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -16,7 +16,7 @@ regex = "1.8.2"
 termcolor = "1.2.0"
 tree-sitter = "0.20.10"
 tree-sitter-javascript = "0.20.0"
-tree-sitter-rust = "0.20.3"
+tree-sitter-rust = { git = "https://github.com/helixbass/tree-sitter-rust", rev = "6327788" }
```
helixbass, line 19: The approach was to fork the tree-sitter grammar and tell it to attempt to parse `(...)`/`[...]`-style macro invocations as a sequence of expressions

I opened a [dummy PR](https://github.com/helixbass/tree-sitter-rust/pull/1)

peterstuart (approved):
Do you know why this isn't already how the grammar works? I thought it might fail if you gave it something  that doesn't look like valid Rust, like

```rust
vec![call_something() something_else];
```

(no comma), but it works fine there too.

PR #46, opened 2023-07-01
Use custom `grep` implementation

helixbass:
In this PR:
- since we were running into insurmountable mismatches between `grep`'s world-view vs ours, replace it with "our own version" (aka copied + pasted, stripped-down and slightly modified versions of stuff from `grep`)
- support the use case that drove this refactor which is handling overlapping matches

Not in this PR:
- I'm assuming CI will fail here on a bunch of Clippy/compiler warnings for eg unused things. I'm assuming what I should do for now is go through and mark them all as `#[allow(dead_code)]` or whatever and then at some point in the future if we think we've exposed all of the relevant stuff (eg other options that are symmetrical to existing `ripgrep` options like context/passthru/invert-match) that this code "supports" and there's still some unused stuff then delete it?
- Given that my starting point was copying + pasting code from `grep` it seems like I should add "attributions" to the top of those source files? And add the same/compatible license that it's distributed under?

To test:
Performance of a release build looks like it's a touch faster than the `grep` version (I was spitting out >8000 match output lines against the Typescript compiler in Rust codebase in ~360ms, pretty cool)
Existing behavior should still work and per the newly added test cases should be able to successfully print out "overlapping" matches as expected (in both "normal" and `--vimgrep` mode)


Cargo.toml
```diff
@@ -6,13 +6,20 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+bstr = "1.1.0"
```
helixbass, line 9: I tried to mimic the exact version of the dependencies of `grep` but some of them look kind of old so might be fine to try and upgrade?

Cargo.toml
```diff
@@ -31,6 +38,9 @@ escargot = "0.5.7"
 predicates = "3.0.3"
 shlex = "1.1.0"
 
+[features]
+default = ["bytecount/runtime-dispatch-simd"]
```
helixbass, line 42: This I also just mimicked as it looked appropriate to do so but don't know exactly what this invokes

src/lib.rs
```diff
@@ -14,26 +14,32 @@ use tree_sitter::Query;
 
 mod args;
 mod language;
+mod line_buffer;
```
helixbass, line 17: I didn't keep the existing `grep_*` crates as separate crates (nor even strictly in subdirectories like they were in), just "hoisted" all of the relevant modules to the top level

src/lib.rs
```diff
@@ -14,26 +14,32 @@ use tree_sitter::Query;
...
 mod macros;
 mod matcher;
 mod plugin;
 mod printer;
 mod project_file_walker;
+mod query_context;
 mod searcher;
+mod sink;
 mod treesitter;
+mod use_printer;
```
helixbass, line 28: ...and so then to avoid a namespace conflict used the (maybe questionable) naming convention of `use_*` for "our configuration/usage of ___"

src/lib.rs
```diff
@@ -127,8 +133,8 @@ pub fn run(args: Args) {
             let path =
                 format_relative_path(project_file_dir_entry.path(), args.is_using_default_paths());
 
-            let matcher = TreeSitterMatcher::new(
-                &query,
+            let query_context = QueryContext::new(
```
helixbass, line 136: A `QueryContext` is basically just bundling up the same things that were passed to a `TreeSitterMatcher`

I got rid of `Matcher`'s entirely

src/line_buffer.rs
```diff
@@ -0,0 +1,32 @@
+use std::io;
```
helixbass, line 1: In general there's probably no easy way to see the "diff" of these vs the corresponding `grep` files

Mostly I stripped out stuff that I decided we didn't want to support/target at least for now (eg binary-file detection (seems questionable to treat that as a "thing" when by definition we're only dealing with parse-able source code files?), the concept of "captures" (seems intriguing that we could actually do something similar as far as tree-sitter query "matches"/"captures" and maybe leverage some of the existing code but I wasn't clear on what that modeling would be so stripped it out for now), "replacing" (`ripgrep` doesn't edit your files, but it does have a "read-only" "search-and-replace" functionality, which seems like it might be nice to support something similar, but from what I could tell it strictly depends on captures in its current form so stripped it out))

And then made a few changes to better "match" our use case

I also stripped out all comments, I like the idea that the corresponding original `grep` code is "source-of-truth". So maybe we could add comments for some of the stuff we modified if we want to. And I guess if we end up wanting to expose any of the `grep`-based stuff as part of a crate-level API then we'd want to re-add the relevant doc comments (to be able to generate docs for those things)?

src/matcher/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    #[inline]
+    pub fn with_end(&self, end: usize) -> Match {
+        assert!(self.start <= end, "{} is not <= {}", self.start, end);
+        Match { end, ..*self }
+    }
+
+    #[inline]
+    pub fn with_end_if_extends(&self, end: usize) -> Match {
```
helixbass, line 47: I added this since it was assuming somewhere that when it was "lumping" multiple matches together into a single range of lines the newer "lumpee"/match would end after the previous one(s) but that invariant wasn't actually true in tree-sitter world

src/matcher/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+
+impl Default for LineTerminator {
+    #[inline]
+    fn default() -> LineTerminator {
+        LineTerminator::byte(b'\n')
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct ByteSet(BitSet);
```
helixbass, line 155: This probably isn't used if it's only used for binary-detection stuff

src/printer/standard.rs
```diff
@@ -0,0 +1,970 @@
...
+        Ok(!self.should_quit())
+    }
+
+    fn context(&mut self, searcher: &Searcher, ctx: &SinkContext<'_>) -> Result<bool, io::Error> {
+        self.standard.matches.clear();
+
+        if ctx.kind() == &SinkContextKind::After {
+            self.after_context_remaining = self.after_context_remaining.saturating_sub(1);
+        }
+        // if searcher.invert_match() {
```
helixbass, line 354: I left this commented out as sort of a "reminder" that I'm pretty sure that if we try and expose an invert-match option it won't "just work" (this is where in that case it was doing the weird "re-matching" to get exact match locations, which now the idea is they are getting passed explicitly to the sink at the time of telling it about a "match" (which may in practice be a "clump of match lines")

src/printer/stats.rs
```diff
@@ -0,0 +1,122 @@
+use std::{
+    ops::{Add, AddAssign},
+    time::Duration,
+};
+
+use super::util::NiceDuration;
+
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+#[cfg_attr(feature = "serde1", derive(serde::Serialize))]
```
helixbass, line 9: Oh I should strip this `serde1` feature attribute, I was attempting to remove its existence (ie make it always compile with the `serde` support)

src/searcher/core.rs
```diff
@@ -0,0 +1,258 @@
...
+    absolute_byte_offset: u64,
+    line_number: Option<u64>,
+    last_line_counted: usize,
+    last_line_visited: usize,
+    after_context_left: usize,
+    has_sunk: bool,
+}
+
+impl<'s, S: Sink> Core<'s, S> {
+    pub fn new(searcher: &'s Searcher, query_context: QueryContext, sink: S) -> Core<'s, S> {
```
helixbass, line 24: Ok so basically a `QueryContext` gets passed mostly everywhere where a `Matcher` was getting passed before, and then in the appropriate place we "inline" use the `QueryContext` to get an iterator over tree-sitter query capture matches (so basically the searcher has tree-sitter-specific stuff hardcoded in)

src/searcher/core.rs
```diff
@@ -0,0 +1,258 @@
...
+        let keepgoing = self.sink.matched(
+            self.searcher,
+            &SinkMatch {
+                line_term: self.config.line_term,
+                bytes: linebuf,
+                absolute_byte_offset: offset,
+                line_number: self.line_number,
+                buffer: buf,
+                bytes_range_in_buffer: range.start()..range.end(),
+                exact_matches,
```
helixbass, line 155: Here's where we're passing the `exact_matches` that we've been keeping track of to the `Sink` API

src/searcher/glue.rs
```diff
@@ -0,0 +1,262 @@
...
+    lines::{self, LineStep},
+    plugin::get_loaded_filter,
+    query_context::QueryContext,
+    searcher::{core::Core, Config, Range, Searcher},
+    sink::Sink,
+    treesitter::{get_parser, node_to_match},
+};
+
+#[derive(Debug, Default)]
+struct AccumulatedExactMatches {
```
helixbass, line 13: This type and the places where we add "exact matches" to it are a little janky-feeling but the basic idea is it looked like we need to keep track of the beginning of the first line of the "clump" of match lines that are going to be the byte range that gets passed to the sink API so that we can specify the exact matches that we're now passing there as relative to that byte range

src/searcher/glue.rs
```diff
@@ -0,0 +1,262 @@
...
+}
+
+impl AsRef<[Range]> for AccumulatedExactMatches {
+    fn as_ref(&self) -> &[Range] {
+        &self.matches_with_offsets_relative_to_reference_beginning_of_line_offset
+    }
+}
+
+#[derive(Debug)]
+pub struct MultiLine<'s, S> {
```
helixbass, line 50: We're always in "multi-line" mode from `grep`'s perspective but I kept most of the existing names the same (eg here it's not very "meaningful" that this is called `MultiLine`, but it is helpful for recognizing its correspondence to the same type in the `grep` source code)

src/searcher/glue.rs
```diff
@@ -0,0 +1,262 @@
...
+            slice,
+            last_match: None,
+            accumulated_exact_matches: Default::default(),
+        }
+    }
+
+    pub fn run(mut self) -> Result<(), S::Error> {
+        if self.core.begin()? {
+            let mut keepgoing = true;
+            let mut query_cursor = QueryCursor::new();
```
helixbass, line 77: Here in our "main searching loop" is where we initialize the tree-sitter query stuff

src/searcher/glue.rs
```diff
@@ -0,0 +1,262 @@
...
+                .filter_map(|(match_, found_capture_index)| {
+                    let found_capture_index = found_capture_index as u32;
+                    if found_capture_index != capture_index {
+                        return None;
+                    }
+                    let mut nodes_for_this_capture = match_.nodes_for_capture_index(capture_index);
+                    let single_captured_node = nodes_for_this_capture.next().unwrap();
+                    assert!(
+                        nodes_for_this_capture.next().is_none(),
+                        "I guess .captures() always wraps up the single capture like this?"
```
helixbass, line 98: This is worth sanity-checking, I found it strange that the "captures-oriented" query API had the same `Iterator::Item` type as the "matches-oriented" one

tests/output.rs
```diff
@@ -675,3 +675,42 @@ fn test_macro_contents() {
         "#,
     );
 }
+
+#[test]
+fn test_sorting_maybe_nesting_related() {
```
helixbass, line 680: So I closed #43 because my understanding was that by switching to the "captures-oriented" query API + maybe some of the custom `grep` stuff we would be guaranteed to not have this ordering issue, but this branch was based on that branch so the fact that the test case from that branch passes (without the explicit `.collect()`'ing + `.sort()`'ing from that branch) is a good sign

src/printer/stats.rs
peterstuart: Does the "1" there just mean "version 1 of serde"?

tests/output.rs
```diff
@@ -675,3 +675,42 @@ fn test_macro_contents() {
...
+
+#[test]
+fn test_overlapping_matches_vimgrep() {
+    assert_sorted_output(
+        "rust_overlapping",
+        r#"
+            $ tree-sitter-grep --query-source '(closure_expression) @closure_expression' --language rust --vimgrep
+            src/lib.rs:2:13:    let f = || {
+            src/lib.rs:3:9:        || {
+        "#,
```
peterstuart, line 714: So is the behavior here that `vimgrep` mode gets you the first line of every match (even if they overlap_, but "normal" mode will "coalesce" overlapping matches?

peterstuart (commented):
I didn't read through all of the copied code, but working + slightly faster sounds good to me 😄 

> I'm assuming CI will fail here on a bunch of Clippy/compiler warnings for eg unused things. I'm assuming what I should do for now is go through and mark them all as #[allow(dead_code)] or whatever and then at some point in the future if we think we've exposed all of the relevant stuff (eg other options that are symmetrical to existing ripgrep options like context/passthru/invert-match) that this code "supports" and there's still some unused stuff then delete it?

Yeah, that sounds reasonable, or just globally allow `dead_code` if you don't want to deal with that?

> Given that my starting point was copying + pasting code from grep it seems like I should add "attributions" to the top of those source files? And add the same/compatible license that it's distributed under?

I'm not really sure what the convention is for that, but that seems reasonable. Adding attributing in a README would probably be sufficient too, but it's nice to be explicit (and not just for license reasons).

src/printer/stats.rs
helixbass: I'm not sure what the "1" means, it looks like that feature controls whether the `serde*` dependencies are included/utilized

Specifically it looks like the JSON printer depends on it

src/treesitter/mod.rs
```diff
@@ -15,29 +14,8 @@ pub(crate) fn maybe_get_query(source: &str, language: Language) -> Option<Query>
...
-        .matches(query, tree.root_node(), file_text_as_bytes)
-        .flat_map(|match_| {
-            match_
-                .nodes_for_capture_index(capture_index)
-                .filter(|node| filter.map_or(true, |filter| filter.call(node)))
-                .collect::<Vec<_>>()
-        })
-        .map(|node| {
-            let range = node.range();
+pub(crate) fn node_to_match(node: &Node) -> Match {
```
helixbass: Updated to express this as a `From` implementation instead

tests/output.rs
helixbass: Ya "normal" mode shows you all the lines of the matching AST nodes (this corresponds to the printer option `per_match_one_line`) and "coalesces" any of those lines that overlap (`per_match`), vs `vimgrep` mode shows you just the first line of each matching AST node and doesn't coalesce them if they overlap (ie if two matches start on the same line)

PR #47, opened 2023-07-02
Expose context-lines options

helixbass:
In this PR:
- expose `--after-context`/`-A`, `--before-context`/`-B`, `--context`/`-C` command-line options mimicking the corresponding `ripgrep` options (for showing "contextual lines" around match lines

Not in this PR:
- `ripgrep` appears to be doing some more sophisticated "resolution" of its supplied command-line flags eg it appears that "order matters" if you specify "competing" flags (like both `--context` + `--after-context`). If we wanted to support that I _believe_ we would have to abandon the `clap`-`#[derive(...)]`-based approach (because it doesn't expose things like order of the command-line arguments to you)? So that seems like it would be a big "choice" to make but I didn't see it as super compelling to jump into that to maintain that level of `ripgrep`-"compatibility" at the moment

To test:
Per newly added test cases you should see surrounding lines around matches being printed basically the way `ripgrep` does it by default (just without color formatting) when you specify one or more (see above caveat) of the context-lines command-line options

Based on `custom-grep`

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -40,6 +40,15 @@ pub struct Args {
 
     #[arg(long)]
     vimgrep: bool,
+
+    #[arg(short = 'A', long, value_name = "NUM")]
+    pub after_context: Option<usize>,
+
+    #[arg(short = 'B', long, value_name = "NUM")]
+    pub before_context: Option<usize>,
+
+    #[arg(short = 'C', long, value_name = "NUM")]
+    pub context: Option<usize>,
 }
 
 impl Args {
@@ -71,9 +80,24 @@ impl Args {
         self.vimgrep
     }
 
+    fn contexts(&self) -> (usize, usize) {
+        let both = self.context.unwrap_or(0);
+        if both > 0 {
+            (both, both)
+        } else {
+            (
+                self.before_context.unwrap_or(0),
+                self.after_context.unwrap_or(0),
+            )
+        }
+    }
+
     pub(crate) fn get_searcher(&self) -> Searcher {
+        let (before_context, after_context) = self.contexts();
         SearcherBuilder::new()
             .line_number(self.line_number())
+            .before_context(before_context)
+            .after_context(after_context)
             .build()
     }
 
--- a/src/args.rs
+++ b/src/args.rs
@@ -1,8 +1,14 @@
 use std::path::{Path, PathBuf};
 
 use clap::{ArgGroup, Parser};
+use termcolor::BufferWriter;
 
-use crate::language::SupportedLanguageName;
+use crate::{
+    language::SupportedLanguageName,
+    printer::StandardBuilder,
+    searcher::{Searcher, SearcherBuilder},
+    use_printer::Printer,
+};
 
 #[derive(Parser)]
 #[clap(group(
@@ -49,17 +55,33 @@ impl Args {
         self.paths.is_empty()
     }
 
-    pub(crate) fn output_mode(&self) -> OutputMode {
-        if self.vimgrep {
-            OutputMode::Vimgrep
-        } else {
-            OutputMode::Normal
-        }
+    fn line_number(&self) -> bool {
+        true
     }
-}
 
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub(crate) enum OutputMode {
-    Normal,
-    Vimgrep,
+    fn per_match(&self) -> bool {
+        self.vimgrep
+    }
+
+    fn per_match_one_line(&self) -> bool {
+        self.vimgrep
+    }
+
+    fn column(&self) -> bool {
+        self.vimgrep
+    }
+
+    pub(crate) fn get_searcher(&self) -> Searcher {
+        SearcherBuilder::new()
+            .line_number(self.line_number())
+            .build()
+    }
+
+    pub(crate) fn get_printer(&self, buffer_writer: &BufferWriter) -> Printer {
+        StandardBuilder::new()
+            .per_match(self.per_match())
+            .per_match_one_line(self.per_match_one_line())
+            .column(self.column())
+            .build(buffer_writer.buffer())
+    }
 }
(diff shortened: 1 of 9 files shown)


src/args.rs
```diff
@@ -49,17 +64,48 @@ impl Args {
         self.paths.is_empty()
     }
 
-    pub(crate) fn output_mode(&self) -> OutputMode {
```
helixbass, line 52: It clicked for me that if (as I'd think we would) we would expose some other output-related "knobs" like `ripgrep` does eg `--heading`, `--column` then an enumerated `OutputMode` is not the right abstraction

So refactored to get rid of `OutputMode` and do something much more similar to what `ripgrep` does (make the `Args` "smarter" and specifically know how to produce configured eg searchers/printers based on its contained settings)

src/args.rs
```diff
@@ -49,17 +64,48 @@ impl Args {
...
-        if self.vimgrep {
-            OutputMode::Vimgrep
+    fn line_number(&self) -> bool {
+        true
+    }
+
+    fn per_match(&self) -> bool {
```
helixbass, line 71: This would basically correspond to a "knob" for "should we print every match (vs congeal adjacent/overlapping match lines)", which seems like as we've discussed would maybe be useful(/expected? Should it be the default?) for our use case

src/args.rs
```diff
@@ -49,17 +64,48 @@ impl Args {
...
+        self.vimgrep
```
helixbass, line 72: So per above now `--vimgrep` is treated more like how `ripgrep` treats it which is almost like a "shorthand" for a bunch of individual knob settings

src/args.rs
```diff
@@ -49,17 +64,48 @@ impl Args {
...
+    }
+
+    fn column(&self) -> bool {
+        self.vimgrep
+    }
+
+    fn contexts(&self) -> (usize, usize) {
+        let both = self.context.unwrap_or(0);
+        if both > 0 {
+            (both, both)
```
helixbass, line 86: This mimics `ripgrep`'s logic but per PR comment, if you currently supplied a combination of `--context` and either `--after-context`/`--before-context`, since we don't do the "last one wins" command-line-option-handling logic, that means that you'd get the maybe unintuitive behavior that `--context` "wins"/"outweighs" the more specific additionally-provided `--after-context`/`--before-context` value

But I'm not too concerned with that at the moment, it's "weird"/"non-optimal" usage

src/use_printer.rs
```diff
@@ -1,46 +1,26 @@
...
-    printer::{Standard, StandardBuilder},
-};
+use crate::{printer::Standard, Args};
 
-type Printer = Standard<Buffer>;
+pub type Printer = Standard<Buffer>;
 
 thread_local! {
-    static PRINTER: OnceCell<(Rc<RefCell<Printer>>, OutputMode)> = Default::default();
+    static PRINTER: OnceCell<(Rc<RefCell<Printer>>, *const Args)> = Default::default();
```
helixbass, line 14: Look at me being clever and avoiding lifetime stuff by storing a `*const Args` here (to maintain the same "invariant sanity-check" as was here previously) and knowing that I won't run into `unsafe` because I'm not actually going to dereference that pointer (just compare it by pointer identity)

tests/output.rs
```diff
@@ -714,3 +717,333 @@ fn test_overlapping_matches_vimgrep() {
...
+
+#[test]
+fn test_after_context_overlapping_multiline_matches_vimgrep() {
+    assert_sorted_output(
+        "rust_overlapping",
+        r#"
+            $ tree-sitter-grep -q '(closure_expression) @c' -l rust --after-context 2 --vimgrep
+            src/lib.rs:2:13:    let f = || {
+            src/lib.rs:3:9:        || {
+            src/lib.rs-7-}
```
helixbass, line 784: This is kind of weird that there's a "gap" between the match lines printed by `--vimgrep` (which only outputs the first line of each match) and then the context lines (which are the lines after the actual end of the matched AST nodes)

But it does in fact appear to match the corresponding `ripgrep` behavior (if you combine a `--multiline` search whose matches do in fact span multiple lines with `--context`/`--after-context` and `--vimgrep` you also see such a "gap")

tests/output.rs
```diff
@@ -714,3 +717,333 @@ fn test_overlapping_matches_vimgrep() {
...
+        r#"
+            $ tree-sitter-grep -q '(function_item) @f' -l rust --after-context 2
+            src/stop.rs:1:fn stop_it() {}
+            src/helpers.rs:1:pub fn helper() {}
+            src/lib.rs:3:pub fn add(left: usize, right: usize) -> usize {
+            src/lib.rs:4:    left + right
+            src/lib.rs:5:}
+            src/lib.rs-6-
+            src/lib.rs-7-#[cfg(test)]
+            --
```
helixbass, line 734: I don't know what the exact "rules" are for when it throws in one of these `--` "context separator lines" but I feel pretty confident that we're getting the same behavior as `grep`/`ripgrep`

tests/output.rs
```diff
@@ -714,3 +717,333 @@ fn test_overlapping_matches_vimgrep() {
...
+            src/lib.rs:5:}
+            src/lib.rs-6-
+            --
+            src/lib.rs-10-
+            src/lib.rs-11-    #[test]
+            src/lib.rs:12:    fn it_works() {
+            src/lib.rs:13:        let result = add(2, 2);
+            src/lib.rs:14:        assert_eq!(result, 4);
+            src/lib.rs:15:    }
+            src/lib.rs-16-}
```
peterstuart, line 1046: I love this whole style of testing.

tests/output.rs
```diff
@@ -127,7 +127,8 @@ fn strip_trailing_carriage_return(line: &str) -> Cow<'_, str> {
 }
 
 fn normalize_match_path(line: &str) -> Cow<'_, str> {
-    regex!(r#"^[^:]+:"#).replace(line, |captures: &Captures| captures[0].replace('\\', "/"))
+    regex!(r#"^[^:]+[:-]\d+[:-]"#)
```
helixbass, line 130: Windows CI was terribly offended that I hadn't updated this regex to now also massage context output lines

PR #48, opened 2023-07-05
Errors for no files searched, invalid paths

helixbass:
In this PR:
- in working on the `ignore` refactor I ran into stuff where it seemed like cleaning up how we're handling/tracking the state of the combination of the `--language` flag, specified path arguments (files and/or directories) and whether we could parse the query would be a good "first step"

To test:
Per additional test cases, when it failed to parse the query for the specified or any of the auto-detected languages (and "nothing else went wrong") you should now see which languages it tried to parse the query in in the error message
When it didn't search any files (because there were no recognized file types in the current directory or specified paths) it should show an error message that it didn't search any files
When you specify a nonexistent directory and/or file path on the command line you should see corresponding file not found error messages


src/lib.rs
```diff
@@ -117,9 +148,12 @@ pub fn run(args: Args) {
     let cached_queries: CachedQueries = Default::default();
     let capture_index = CaptureIndex::default();
     let buffer_writer = BufferWriter::stdout(ColorChoice::Never);
+    let matched = AtomicBool::new(false);
```
helixbass, line 151: This basically mimics `ripgrep`

src/lib.rs
```diff
@@ -140,19 +174,41 @@ pub fn run(args: Args) {
...
+    if !messages::errored() {
+        if !searched.load(Ordering::SeqCst) {
+            eprint_nothing_searched();
+        } else {
+            cached_queries.error_if_no_successful_query_parsing();
+        }
+    }
+
+    if messages::errored() {
+        process::exit(2);
```
helixbass, line 197: This mimics `ripgrep`, I guess it uses exit status 2 to indicate "actual failure" and 1 to indicate "no matches found"

src/project_file_walker.rs
```diff
@@ -34,7 +37,10 @@ impl WalkParallelIterator {
                     let sender = sender.clone();
                     move |entry| {
                         let entry = match entry {
-                            Err(_) => return WalkState::Continue,
+                            Err(err) => {
+                                err_message!("{err}");
```
helixbass, line 41: This is where the file not found error is surfaced from `WalkParallel` (for both directory and file path command-line arguments)

tests/output.rs
```diff
@@ -1048,3 +1067,110 @@ fn test_before_and_after_context() {
...
+    assert_failure_output(
+        "rust_project",
+        r#"
+            $ tree-sitter-grep -q '(function_item) @f' srcz/
+            srcz/: No such file or directory (os error 2)
+        "#,
+    );
+}
+
+// #[test]
```
helixbass, line 1151: These test cases I believe will all be able to be addressed as part of the `ignore` refactor PR (since they rely on being able to know which file-type a project-file-walker file-path is associated with)

src/lib.rs
```diff
@@ -68,6 +72,19 @@ impl CaptureIndex {
...
 const ALL_NODES_QUERY: &str = "(_) @node";
 
+fn join_with_or<TItem: fmt::Display>(list: &[TItem]) -> String {
+    let mut ret: String = Default::default();
+    for (index, item) in list.iter().enumerate() {
+        ret.push_str(&item.to_string());
+        if list.len() >= 2 && index < list.len() - 2 {
+            ret.push_str(", ");
+        } else if list.len() >= 2 && index == list.len() - 2 {
+            ret.push_str(" or ");
```
peterstuart, line 82: I vote for an Oxford comma (?): "`, or " 😄

src/lib.rs
peterstuart: Maybe worth making an `ExitCode` enum (and then I guess your own `exit` function) to make that clear?

tests/output.rs
```diff
@@ -346,7 +368,7 @@ fn test_invalid_query_inline() {
         "rust_project",
         r#"
             $ tree-sitter-grep --query-source '(function_itemz) @function_item' --language rust
-            error: invalid query
+            error: couldn't parse query for Rust
```
peterstuart, line 371: Does tree-sitter give any meaningful error messages when trying to parse a query that could be shown here?

peterstuart (approved):

src/lib.rs
helixbass: Sure

I special-cased the `list.len() == 2` case because I assume you wouldn't say eg "Typescript, or Javascript" (instead in that case "Typescript or Javascript")?

Oh also if we want to get even more posh I slightly considered using "nor" instead of "or" here

tests/output.rs
helixbass: Ok ya that's a good idea, I made `CachedQueries`'s `HashMap` contain a `OnceLock<Result<...>>` instead of a `OnceLock<Option<...>>` but then its `.get_and_cache_query_for_language()` "API" still returns an `Option<...>` (ie it captures the query-parsing error but keeps that encapsulated)

And so then when it kicks in this "no successfully parsed queries behavior", if it only tried parsing for one language then it shows the error message from `tree-sitter` (figured that showing n query-parsing error messages would probably be "a lot" in the tried-to-parse-as-more-than-one-language case?)

PR #49, opened 2023-07-06
Use ignore match metadata

helixbass:
In this PR:
- use a forked version of `ignore` (https://github.com/helixbass/ripgrep/pull/1) which exposes "match metadata" of which specified file-types a given project file that `ignore` returns matched

Not in this PR:
- I haven't added any tests in `ignore` for the newly exposed functionality, I'd like to treat that as a separate task

To test:
Existing behavior should still work and now per newly added test cases you should be able to get different error messages around eg specifying a file path whose type disagrees with the language you specified

Based on `error-invalid-paths`

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -1,11 +1,16 @@
 use std::path::{Path, PathBuf};
 
 use clap::{ArgGroup, Parser};
+use ignore::{types::Types, WalkBuilder, WalkParallel};
+use rayon::iter::IterBridge;
 use termcolor::BufferWriter;
 
 use crate::{
-    language::SupportedLanguageName,
+    language::{SupportedLanguage, SupportedLanguageName},
     printer::StandardBuilder,
+    project_file_walker::{
+        get_project_file_walker_types, into_parallel_iterator, WalkParallelIterator,
+    },
     searcher::{Searcher, SearcherBuilder},
     use_printer::Printer,
 };
@@ -52,7 +57,7 @@ pub struct Args {
 }
 
 impl Args {
-    pub(crate) fn use_paths(&self) -> Vec<PathBuf> {
+    fn use_paths(&self) -> Vec<PathBuf> {
         if self.paths.is_empty() {
             vec![Path::new("./").to_owned()]
         } else {
@@ -107,5 +112,28 @@ impl Args {
             .per_match_one_line(self.per_match_one_line())
             .column(self.column())
             .build(buffer_writer.buffer())
+    }
+
+    pub(crate) fn language(&self) -> Option<SupportedLanguage> {
+        self.language.map(|language| language.get_language())
+    }
+
+    pub(crate) fn get_project_file_walker_types(&self) -> Types {
+        get_project_file_walker_types(self.language())
+    }
+
+    pub(crate) fn get_project_file_walker(&self) -> WalkParallel {
+        let paths = self.use_paths();
+        assert!(!paths.is_empty());
+        let mut builder = WalkBuilder::new(&paths[0]);
+        builder.types(self.get_project_file_walker_types());
+        for path in &paths[1..] {
+            builder.add(path);
+        }
+        builder.build_parallel()
+    }
+
+    pub(crate) fn get_project_file_parallel_iterator(&self) -> IterBridge<WalkParallelIterator> {
+        into_parallel_iterator(self.get_project_file_walker())
     }
 }
--- a/src/args.rs
+++ b/src/args.rs
@@ -35,7 +35,7 @@ pub struct Args {
     pub capture_name: Option<String>,
 
     #[arg(short, long, value_enum)]
-    pub language: Option<SupportedLanguageName>,
+    language: Option<SupportedLanguageName>,
 
     #[arg(short, long)]
     pub filter: Option<String>,
--- a/src/language.rs
+++ b/src/language.rs
@@ -1,6 +1,7 @@
-use std::{collections::HashMap, ffi::OsStr, path::Path};
+use std::collections::HashMap;
 
 use clap::ValueEnum;
+use once_cell::sync::Lazy;
 use tree_sitter::Language;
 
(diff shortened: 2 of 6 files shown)


src/args.rs
```diff
@@ -108,4 +113,27 @@ impl Args {
             .column(self.column())
             .build(buffer_writer.buffer())
     }
+
+    pub(crate) fn language(&self) -> Option<SupportedLanguage> {
```
helixbass, line 117: Went further in the "methods on `Args`" direction, all of these effectively just got moved here

peterstuart: Could make the `language` field private then? (Maybe? It doesn't seem like a crazy thing to have public, but it's not used anywhere else.)

peterstuart: Actually, you could probably have clap directly support `SupportedLanguage` by implementing `FromStr` (or whatever the trait was) on it?

src/language.rs
```diff
@@ -11,115 +12,65 @@ pub enum SupportedLanguageName {
...
+        SupportedLanguage {
+            language: tree_sitter_typescript::language_tsx(),
+            name: SupportedLanguageName::Typescript,
+            name_for_ignore_select: "ts",
+        },
+        SupportedLanguage {
+            language: tree_sitter_javascript::language(),
+            name: SupportedLanguageName::Javascript,
+            name_for_ignore_select: "js",
+        },
```
peterstuart, line 51: This seems pretty macro-able if you want to eventually add all the languages and not type this out in such a verbose way.

src/lib.rs
```diff
@@ -144,20 +143,66 @@ pub fn run(args: Args) {
...
+                            language,
+                        );
+                    }
+                    None => {
+                        error_explicit_path_argument_not_of_known_type(&project_file_dir_entry);
+                    }
+                }
+                return;
+            }
+            let language = return_if_none!(if matched_languages.len() > 1
```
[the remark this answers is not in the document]
peterstuart: I wonder if `return_if_none` could (optionally?) support using braces instead of parentheses? I was a little confused about the syntax here, because I was really not expecting ~30 lines inside of the parentheses.

src/lib.rs
```diff
@@ -206,6 +252,56 @@ fn eprint_nothing_searched() {
...
+
+#[macro_export]
+macro_rules! only_run_once {
+    ($block:block) => {
+        static ONCE_LOCK: std::sync::OnceLock<()> = OnceLock::new();
+        ONCE_LOCK.get_or_init(|| {
+            $block;
+        });
+    };
+}
```
peterstuart, line 285: Ha, nice, I like this

peterstuart (commented):
I'm having trouble keeping up with all the changes here, but it looks reasonable to me. 🤷🏻

src/lib.rs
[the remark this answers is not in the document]
helixbass: Ok actually you get that "for free" (you can invoke any declarative macro with any "brace type") afaik, and that's a nice idea here (to use `return_if_none! { ... }` presumably)

But per my comment the way this was written was questionable so I restructured the logic here (+ the preceding `if matched_languages.is_empty()` "special case") into more of a nested-`match` style which got rid of the need for `return_if_none!()` at all here

And actually by rewriting it became clearer (if I did want to still be silly about avoiding allocating n more small `Vec`'s) that in fact it was easy/natural to not do that and only allocate that extra `Vec` in the case that there are multiple matching languages (in auto-language mode) (at which point allocating that `Vec` makes the code for that case much simpler)

src/language.rs
helixbass: Or what about just a typical `SupportedLanguage::new()` constructor?

Or were you picturing stripping it down even more than that eg omitting the `SupportedLanguageName::` prefix (which I guess you could achieve via a `use SupportedLanguageName::*;` but could also achieve via an un-hygienic macro (?))?

src/project_file_walker.rs
```diff
@@ -1,52 +1,75 @@
...
+                                    })
+                            })
+                            .map(|match_metadata| {
+                                match_metadata
+                                    .matching_file_types
+                                    .map(|file_type_def| {
+                                        ALL_SUPPORTED_LANGUAGES_BY_NAME_FOR_IGNORE_SELECT
+                                            .get(&file_type_def.name())
+                                            .copied()
+                                            .unwrap()
```
helixbass, line 67: Thinking of eg the Typescript `Record` type (with a union-of-string-literals as its key type), it seems like it might be nice to have it encoded at the type level that lookups into a "known/enumerated keys + fully filled on initialization" `HashMap` like this should never fail (so then consumer-wise you wouldn't `.unwrap()` after `.get()`'ing)?

src/args.rs
peterstuart: Ah yeah, I didn’t think about the help text. I don't see an obvious way to do that in clap without an enum.

It does look like it could be an enum fairly easily. I guess there are some (probably not important) performance things to consider there since it'd be less easy to make sure the per language code only really runs once? (Which is handled in the setup of the SupportedLanguage now?)

src/lib.rs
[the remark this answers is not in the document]
peterstuart: Oh that's interesting. So that's true even for things like `vec![]`?

src/project_file_walker.rs
peterstuart: Hmm, yeah, I'm not sure how you'd achieve that exactly.

src/lib.rs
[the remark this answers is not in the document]
helixbass: Yup it's just [by convention](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c3a1c885281ba6ce107cb891376fd29f) that we always use `[...]` there

src/language.rs
peterstuart: Yeah, that's what I was thinking, assuming you always use the same names as tree sitter does internally, but that's probably not worth it over just using a constructor.

PR #50, opened 2023-07-06
More languages

### LINKED ISSUE #14: Support other languages
I think the plan for an "initial target" is to bundle a bunch more existing tree-sitter language grammar crates and wire them all up to the command-line language flag

helixbass:
In this PR:
- add a bunch of other languages

Closes #14 

It looks like the release binary size went from ~6M to ~24M with these additional language crates added, I guess that was more or less to be expected and doesn't seem that large for a binary executable?

To test:
If you want to try running queries against any of the newly added languages they should work as expected/symmetrically to existing behavior
You should be able to get a "`*.h` is ambiguous" error if you pass a query that could be parsed as more than one of a C/C++/Objective-C query in auto-language mode in a directory containing a `*.h` file
Conversely, if you pass a query that eg would only parse for Objective-C in the above scenario, it should run that query as an Objective-C query against the `*.h` files

Based on `ignore-match-metadata`


Cargo.toml
```diff
@@ -12,7 +12,7 @@ bytecount = "0.6"
 clap = { version = "4.3.0", features = ["derive"] }
 encoding_rs = "0.8.14"
 encoding_rs_io = "0.1.6"
-ignore = { git = "https://github.com/helixbass/ripgrep", rev = "e811564" }
+ignore = { git = "https://github.com/helixbass/ripgrep", rev = "012f2f1" }
```
helixbass, line 15: The existing known file-types in `ignore` only included `*.scm` under `"lisp"` (whose globs overlapped with Elisp) so I added an explicit file-type for tree-sitter-queries to avoid that being an issue

tests/output.rs
```diff
@@ -407,7 +407,7 @@ fn test_invalid_language_name() {
         r#"
             $ tree-sitter-grep --query-source '(function_item) @function_item' --language rustz
             error: invalid value 'rustz' for '--language <LANGUAGE>'
-              [possible values: rust, typescript, javascript]
+              [possible values: rust, typescript, javascript, swift, objective-c, toml, python, ruby, c, cpp, go, java, c-sharp, kotlin, elisp, elm, dockerfile, html, tree-sitter-query, json, css, lua]
```
peterstuart, line 410: Maybe these could be alphabetized to feel a bit less randomly-ordered?

tests/output.rs
```diff
@@ -1180,3 +1191,440 @@ fn test_specify_explicit_file_of_unrecognized_file_type_and_language_flag() {
...
+    );
+}
+
+#[test]
+fn test_objective_c_auto_language_ambiguous_query() {
+    assert_failure_output(
+        "objective_c_project",
+        r#"
+            $ tree-sitter-grep -q '(identifier) @c'
+            File "./example.h" has ambiguous file-type, could be ObjectiveC, C or Cpp. Try passing the --language flag
```
peterstuart, line 1245: I think of "Cpp" as the name you use for C++ when you're in a context where using "+" would be annoying/impossible, but not really as the "real" name of the language. (Although I suppose you'd have to quote "c++" to pass it to `--language`, so maybe being consistent here is best?)

peterstuart (approved):
That seems ok as a binary size (although larger than I'd have guessed?), but if want this to be usable as a crate as well eventually, maybe worth making Cargo features for all the languages, so you can use the crate without needed to get every one of the underlying tree-sitter libraries?

tests/output.rs
```diff
@@ -1180,3 +1191,440 @@ fn test_specify_explicit_file_of_unrecognized_file_type_and_language_flag() {
...
+    );
+}
+
+#[test]
+fn test_objective_c_auto_language_ambiguous_query() {
+    assert_failure_output(
+        "objective_c_project",
+        r#"
+            $ tree-sitter-grep -q '(identifier) @c'
+            File "./example.h" has ambiguous file-type, could be ObjectiveC, C, or Cpp. Try passing the --language flag
```
helixbass, line 1245: > I think of "Cpp" as the name you use for C++ when you're in a context where using "+" would be annoying/impossible, but not really as the "real" name of the language. (Although I suppose you'd have to quote "c++" to pass it to `--language`, so maybe being consistent here is best?)

Good point, and actually I don't think the `+`'s are eg problematic to the shell when passing `--language c++` on the command-line, so updated

Specifically `#[derive(ValueEnum)]` supports an eg `#[value(name = "c++")]` declarative attribute

Where that got interesting is that as of #53 that enum definition is macro-generated

So I updated the macro code to preserve such an attribute (I was already parsing that chunk as a `syn::ExprArray` which "captured" the attribute correctly, I was just then ignoring/stripping it out)

helixbass (commented):
> if want this to be usable as a crate as well eventually, maybe worth making Cargo features for all the languages, so you can use the crate without needed to get every one of the underlying tree-sitter libraries?

That makes sense based on my limited understanding

But could wait to do that until we are actually exposing a crate-level API?

Seems probably worth wrapping our heads around how dynamic loading of languages/grammars might work too as part of/before doing that

PR #51, opened 2023-07-07
Only-matching option

helixbass:
In this PR:
- expose `--only-matching`/`-o` option like `ripgrep` for only outputting the parts of lines that actually correspond to what was matched (vs always outputting entire lines)

To test:
Per newly added test cases, if you use `--only-matching`/`-o` and your matches don't constitute all of the text on all of the lines that they're on, you should only see the text corresponding to the matched tree-sitter AST node in the output

Based on `more-languages`


tests/languages.rs
```diff
@@ -0,0 +1,440 @@
+mod shared;
```
helixbass, line 1: I split the tests into multiple files since the one file was getting pretty long

From what I could tell in order to accomplish that you have to make them new separate "top-level" integration tests (under `tests/`)? Which seems fine

And per some guides extracted the shared stuff into a shared module that also sits under `tests/`

tests/output.rs
```diff
@@ -1193,438 +980,54 @@ fn test_specify_explicit_file_of_unrecognized_file_type_and_language_flag() {
...
-        r#"
-            $ tree-sitter-grep -q '(identifier) @c'
-            File "./example.h" has ambiguous file-type, could be ObjectiveC, C or Cpp. Try passing the --language flag
-        "#,
-    );
-}
-
-#[test]
-fn test_toml() {
+fn test_only_matching() {
```
helixbass, line 983: Here are the newly added tests

tests/output.rs
```diff
@@ -1193,438 +980,54 @@ fn test_specify_explicit_file_of_unrecognized_file_type_and_language_flag() {
...
     assert_sorted_output(
         "rust_project",
         r#"
-            $ tree-sitter-grep -q '(string) @c' --language toml
-            Cargo.toml:2:name = "rust_project"
-            Cargo.toml:3:version = "0.1.0"
-            Cargo.toml:4:edition = "2021"
+            $ tree-sitter-grep --query-source '(parameter) @c' --language rust --only-matching
+            src/lib.rs:3:left: usize
+            src/lib.rs:3:right: usize
```
helixbass, line 989: Since it looks like the underlying printer (and per `ripgrep`'s docs) prints two separate output lines if you do `--only-matching` and there are multiple matches found on the same line (I guess since otherwise it might be printing non-match text between the two matches on that line?), I was curious how it would handle cases with eg multi-line/overlapping matches, so added a couple tests for that below. What it did seems reasonable/not "broken"

tests/output.rs
helixbass: Agreed not good default behavior/"prioritization"

I just ran into this right now when trying to search for/jump to this comment in the Files tab (after having read it in the Conversation tab - that's my workflow because I don't think it lets you aggregate comments as a "review" if you add those comments from the Conversation tab?)

tests/output.rs
helixbass: Ya unless we're opinionated I think that "mimicking a lot of what `ripgrep` does (by leveraging the same `grep`/`ignore` options that it is)" is a good idea/"good enough"

tests/output.rs
```diff
@@ -1193,438 +980,54 @@ fn test_specify_explicit_file_of_unrecognized_file_type_and_language_flag() {
...
-        "lua_project",
+        "rust_overlapping_start_same_line",
         r#"
-            $ tree-sitter-grep -q '(identifier) @c'
-            example.lua:1:function hello()
+            $ tree-sitter-grep -q '(closure_expression) @c' -l rust --only-matching
+            src/lib.rs:3:|| { || {
+            src/lib.rs:4:            println!("whee");
+            src/lib.rs:5:        }
+            src/lib.rs:6:    }
```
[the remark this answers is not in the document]
helixbass: Ya it seems like in theory if you invoked "per-match" mode (eg via `--vimgrep`) then you should see them printed separately (whether just their first lines (or I suppose in this case just the matching part of their first lines) or all of their lines (or again in this case just the actual matching text))?

But that seems like it's poking deep into `ripgrep` etc's "model" to where I'm not necessarily even that intrigued to see how it currently handles some "combinations" like that (but we certainly could)

PR #52, opened 2023-07-09
Exploration: supported languages as "singleton values"

helixbass:
In this PR:
- explore the idea of using fixed-size collections to represent the "all supported languages" collection data structures

Based on `more-languages`

### DIFF
--- a/src/language.rs
+++ b/src/language.rs
@@ -1,11 +1,11 @@
-use std::collections::HashMap;
+use std::{collections::HashMap, ops::Index};
 
 use clap::ValueEnum;
 use once_cell::sync::Lazy;
 use tree_sitter::Language;
 
-#[derive(Copy, Clone, Debug, Eq, PartialEq, ValueEnum, Hash)]
-pub enum SupportedLanguageName {
+#[derive(Copy, Clone, Debug, Eq, PartialEq, ValueEnum)]
+pub enum SupportedLanguage {
     Rust,
     Typescript,
     Javascript,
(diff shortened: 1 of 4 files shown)


src/language.rs
```diff
@@ -1,11 +1,11 @@
...
+use std::{collections::HashMap, ops::Index};
 
 use clap::ValueEnum;
 use once_cell::sync::Lazy;
 use tree_sitter::Language;
 
-#[derive(Copy, Clone, Debug, Eq, PartialEq, ValueEnum, Hash)]
-pub enum SupportedLanguageName {
+#[derive(Copy, Clone, Debug, Eq, PartialEq, ValueEnum)]
+pub enum SupportedLanguage {
```
helixbass, line 8: Ok so probably the idea would be that most of the stuff in this file would get generated by some kind of macro "spec"

`SupportedLanguage` is kind of a "token", just a simple enum that gets used as an index into all of the "all of the supported languages" collections

src/language.rs
```diff
@@ -30,161 +30,148 @@ pub enum SupportedLanguageName {
...
 }
 
-impl SupportedLanguageName {
-    pub fn get_language(&self) -> SupportedLanguage {
-        *ALL_SUPPORTED_LANGUAGES_BY_SUPPORTED_LANGUAGE_NAME
-            .get(self)
-            .unwrap()
-    }
-}
+pub type BySupportedLanguage<T> = [T; 22];
```
helixbass, line 33: Use fixed-size arrays as the actual data structure for the "all supported languages" collections

src/language.rs
```diff
@@ -30,161 +30,148 @@ pub enum SupportedLanguageName {
...
+impl<T> Index<SupportedLanguage> for BySupportedLanguage<T> {
+    type Output = T;
 
-#[derive(Copy, Clone, Debug)]
-pub struct SupportedLanguage {
-    pub language: Language,
-    pub name: SupportedLanguageName,
-    pub name_for_ignore_select: &'static str,
+    fn index(&self, index: SupportedLanguage) -> &Self::Output {
+        &self[index as usize]
```
helixbass, line 39: Basically encapsulate some enum <-> `usize` massaging

src/language.rs
```diff
@@ -30,161 +30,148 @@ pub enum SupportedLanguageName {
...
+            value if value == Self::Kotlin as usize => Self::Kotlin,
+            value if value == Self::Elisp as usize => Self::Elisp,
+            value if value == Self::Elm as usize => Self::Elm,
+            value if value == Self::Dockerfile as usize => Self::Dockerfile,
+            value if value == Self::Html as usize => Self::Html,
+            value if value == Self::TreeSitterQuery as usize => Self::TreeSitterQuery,
+            value if value == Self::Json as usize => Self::Json,
+            value if value == Self::Css as usize => Self::Css,
+            value if value == Self::Lua as usize => Self::Lua,
+            _ => unreachable!(),
```
helixbass, line 68: Obviously this is not technically true (that no `usize` argument could trigger this code path) but I think we'd be ok panicking if we were trying to convert a non-"known good" `usize` back into the enum, so using `From` (instead of eg `TryFrom`) seems more ergonomically desirable

src/language.rs
```diff
@@ -30,161 +30,148 @@ pub enum SupportedLanguageName {
...
+        tree_sitter_c_sharp::language(),
+        tree_sitter_kotlin::language(),
+        tree_sitter_elisp::language(),
+        tree_sitter_elm::language(),
+        tree_sitter_dockerfile::language(),
+        tree_sitter_html::language(),
+        tree_sitter_query::language(),
+        tree_sitter_json::language(),
+        tree_sitter_css::language(),
+        tree_sitter_lua::language(),
```
helixbass, line 124: This would obviously be sketchy as far as not messing up the order if the idea wasn't that it would be macro-generated by something where "the relevant things are grouped together"

I guess I decided to make these + the `name_for_ignore_select`'s their own separate data structures (vs in the existing code they are fields on the `SupportedLanguage` struct) since that seemed like the simplest way to "free up" `SupportedLanguage` to just be a simple "token" enum

But might be interesting to explore how you'd represent the "main type that's getting passed around" (eg `SupportedLanguage`) as something with a uniform set of fields too

src/language.rs
```diff
@@ -30,161 +30,148 @@ pub enum SupportedLanguageName {
...
+];
+
+#[inline(always)]
+pub fn get_supported_language_name_for_ignore_select(
+    supported_language: SupportedLanguage,
+) -> &'static str {
+    SUPPORTED_LANGUAGE_NAMES_FOR_IGNORE_SELECT[supported_language]
+}
+
 pub static ALL_SUPPORTED_LANGUAGES_BY_NAME_FOR_IGNORE_SELECT: Lazy<
```
helixbass, line 165: I think this is still "its own separate thing", it's sort of a "special case" I guess where we have a usage pattern where we need to do a "reverse mapping" from some "non-primary" thing back to the main "token" type

src/lib.rs
```diff
@@ -84,25 +83,27 @@ fn join_with_or<TItem: fmt::Display>(list: &[TItem]) -> String {
     ret
 }
 
-struct CachedQueries(HashMap<SupportedLanguageName, OnceLock<Result<Arc<Query>, QueryError>>>);
+#[derive(Default)]
+struct CachedQueries(BySupportedLanguage<OnceLock<Result<Arc<Query>, QueryError>>>);
```
helixbass, line 87: I believe this could now be `static` if we wanted it to be since it looks like `OnceLock::new()` is `const`

But now at least we can `#[derive(Default)]` which will stack-allocate a fixed-size data structure (specifically an array)

src/lib.rs
```diff
@@ -84,25 +83,27 @@ fn join_with_or<TItem: fmt::Display>(list: &[TItem]) -> String {
...
     fn get_and_cache_query_for_language(
         &self,
         query_source: &str,
         language: SupportedLanguage,
     ) -> Option<Arc<Query>> {
-        self.0
-            .get(&language.name)
-            .unwrap()
-            .get_or_init(|| maybe_get_query(query_source, language.language).map(Arc::new))
+        self.0[language]
```
helixbass, line 95: So I guess this was one of the goals, getting rid of the "fallible lookup" (ie the `.unwrap()`)

src/lib.rs
```diff
@@ -111,23 +112,25 @@ impl CachedQueries {
             let attempted_parsings = self
                 .0
                 .iter()
+                .enumerate()
                 .filter(|(_, value)| value.get().is_some())
+                .map(|(index, value)| -> (SupportedLanguage, _) { (index.into(), value) })
```
helixbass, line 117: `usize` <-> enum massaging

src/project_file_walker.rs
```diff
@@ -94,10 +95,10 @@ pub(crate) fn get_project_file_walker_types(language: Option<SupportedLanguage>)
     let mut types_builder = TypesBuilder::new();
     types_builder.add_defaults();
     if let Some(language) = language {
-        types_builder.select(language.name_for_ignore_select);
+        types_builder.select(get_supported_language_name_for_ignore_select(language));
     } else {
-        for language in ALL_SUPPORTED_LANGUAGES.iter() {
-            types_builder.select(language.name_for_ignore_select);
+        for name_for_ignore_select in SUPPORTED_LANGUAGE_NAMES_FOR_IGNORE_SELECT.iter() {
```
helixbass, line 100: Sort of interesting that with the approach of "moving all of these other fields into their own separate fixed-size data structures" now this usage pattern becomes an iteration over "the data structure that just has that one relevant thing in it"

src/language.rs
```diff
@@ -30,161 +30,148 @@ pub enum SupportedLanguageName {
...
-            language: tree_sitter_css::language(),
-            name: SupportedLanguageName::Css,
-            name_for_ignore_select: "css",
-        },
-        SupportedLanguage {
-            language: tree_sitter_lua::language(),
-            name: SupportedLanguageName::Lua,
-            name_for_ignore_select: "lua",
-        },
+pub static ALL_SUPPORTED_LANGUAGES: BySupportedLanguage<SupportedLanguage> = {
```
helixbass, line 73: This one was able to be truly `static`...

src/language.rs
```diff
@@ -30,161 +30,148 @@ pub enum SupportedLanguageName {
...
+        Dockerfile,
+        Html,
+        TreeSitterQuery,
+        Json,
+        Css,
+        Lua,
+    ]
+};
+
+static SUPPORTED_LANGUAGE_LANGUAGES: Lazy<BySupportedLanguage<Language>> = Lazy::new(|| {
```
helixbass, line 101: ...but this one needed a `Lazy` wrapper since I guess `tree_sitter::Language`'s aren't `const`?

peterstuart (approved):
Nice, this seems like a good approach (assuming that some macros were making it all “safe”).

PR #53, opened 2023-07-11
Exploration: supported languages as "singleton values": macros

helixbass:
In this PR:
- add macros for "safely" generating fixed-size "all supported languages" collections
- a couple ergonomic improvements related to these per comments on #52 

To test:
You should see the `fixed_map!()` macro fail if you try to omit any keys, pass any spurious key/values or not pass corresponding values of the expected type
You should see the `by_supported_language!()` macro fail if you omit any supported languages or pass any unknown keys

Based on `supported-languages-fixed-size`


proc_macros/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "proc_macros"
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+Inflector = "0.11.4"
```
helixbass, line 10: A capitalized crate name?

Almost seems ironic since what this crate is doing is case conversion

proc_macros/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+}
+
+fn get_collection_type_definition(
+    collection_type_name: &Ident,
+    variants: &[Ident],
+) -> proc_macro2::TokenStream {
+    let len = variants.len();
+    quote! {
+        #[derive(Default)]
+        pub struct #collection_type_name<T>([T; #len]);
```
helixbass, line 121: Per https://github.com/helixbass/tree-sitter-grep/pull/52#discussion_r1259863355, add a newtype wrapper around the fixed-size collection type (eg `BySupportedLanguage`) for the purpose of exposing "`HashMap`-ish" iteration methods

I actually made the changes corresponding to those comments "statically" before starting the macro stuff so if you'd rather see a pre-macroization version of them you could look at the diff as of https://github.com/helixbass/tree-sitter-grep/pull/53/commits/f672cca2e2646b690818e2ce7dac1dcdb6fe3f1e

proc_macros/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+    }
+}
+
+fn get_iter_implementations(
```
helixbass, line 125: Per above comment this is implementing "`HashMap`-style" `.iter()`/`.values()` on eg `BySupportedLanguage`

proc_macros/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+    }
+}
+
+fn get_deref_implementation(
+    collection_type_name: &Ident,
+    variants: &[Ident],
+) -> proc_macro2::TokenStream {
+    let len = variants.len();
+    quote! {
+        impl<T> Deref for #collection_type_name<T> {
```
helixbass, line 234: Added a `Deref` implementation along with the newtype wrapper (I think it's currently only getting used to call the underlying `.len()`)

proc_macros/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+    }
+}
+
+fn get_collection_generator_macro_definition(
+    collection_type_name: &Ident,
+    variants: &[Ident],
+) -> proc_macro2::TokenStream {
+    let macro_name = format_ident!("{}", collection_type_name.to_string().to_snake_case());
+    quote! {
+        #[macro_export]
```
helixbass, line 279: Ok so this is sort of funny, the way that I was able to get a "generated" eg `by_supported_language!()` macro (that needed to be basically implemented as a proc macro, but per https://github.com/helixbass/tree-sitter-grep/pull/52#discussion_r1259916094 I don't think you could easily just "generate and use a new proc macro"?) was to generate it as a _declarative_ macro which basically delegates to an existing "static" `by_fixed_map!()` proc macro but has the things specific to "this fixed map" (ie the variant names and collection type name eg `BySupportedLanguage`) "pre-curried" as arguments to it (by hardcoding them into the declarative macro body basically)

It's sort of funny that I believe those pre-curried arguments just get passed along from one to the other "as token trees", so they can eg reference non-existing identifiers

proc_macros/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let value_mapping_content;
+        bracketed!(value_mapping_content in input);
+        let mut value_mapping: HashMap<Ident, Expr> = Default::default();
+        while !value_mapping_content.is_empty() {
+            let key: Ident = value_mapping_content.parse()?;
+            value_mapping_content.parse::<Token![=>]>()?;
+            let value: Expr = value_mapping_content.parse()?;
+            value_mapping.insert(key, value);
+            if value_mapping_content.is_empty() {
```
helixbass, line 310: On one of these macros I think I required a trailing comma but there are certainly ways to avoid that eg I copied this pattern from the `syn::ExprArray` [`Parse` implementation](https://docs.rs/syn/latest/src/syn/expr.rs.html#1855-1877)

proc_macros/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+#[proc_macro]
+pub fn by_fixed_map(input: TokenStream) -> TokenStream {
+    let ByFixedMapArgs {
+        value_mapping,
+        variants,
+        collection_type_name,
+    } = parse_macro_input!(input as ByFixedMapArgs);
+
+    if value_mapping.len() != variants.len() {
+        panic!("Incorrect variants");
```
helixbass, line 336: These could be more helpful and tell you eg which variant(s) were missing or what invalid variant names it encountered

src/language.rs
```diff
@@ -95,81 +32,79 @@ pub static ALL_SUPPORTED_LANGUAGES: BySupportedLanguage<SupportedLanguage> = {
...
-        tree_sitter_dockerfile::language(),
-        tree_sitter_html::language(),
-        tree_sitter_query::language(),
-        tree_sitter_json::language(),
-        tree_sitter_css::language(),
-        tree_sitter_lua::language(),
-    ]
-});
+impl SupportedLanguage {
+    pub fn language(&self) -> Language {
```
helixbass, line 39: Per https://github.com/helixbass/tree-sitter-grep/pull/52#discussion_r1259873373, move these onto methods on `SupportedLanguage`

src/language.rs
```diff
@@ -95,81 +32,79 @@ pub static ALL_SUPPORTED_LANGUAGES: BySupportedLanguage<SupportedLanguage> = {
...
-    "elm",
-    "docker",
-    "html",
-    "treesitterquery",
-    "json",
-    "css",
-    "lua",
-];
+static SUPPORTED_LANGUAGE_LANGUAGES: Lazy<BySupportedLanguage<Language>> = Lazy::new(|| {
+    by_supported_language!(
```
helixbass, line 49: I think I sort of like this approach of exposing a secondary macro for allowing you to create new "fixed-size collections" (as opposed to eg specifying all of the `tree_sitter::Language`'s and "names for ignore select" in the body of the main `fixed_map!()` invocation and then either generating the same thing as this under the hood or possible generating "fat `SupportedLanguage` structs" (like how it used to have `language`/`name_for_ignore_select` as struct fields rather than these separate top-level data structures))?

I guess the downsides could be (a) in terms of code readability if one thought it were clearer to read "one big declaration" of all of the info/"fields" for all of the supported languages and (b) in terms of runtime if for some reason it were more performant ("cache locality"...?) (and if this were then actually a practical performance issue, I'm really just theorizing) to actually have those fields co-located as part of their relevant "`SupportedLanguage` struct" (vs being co-located with other instances of "that same field" like this)?

src/project_file_walker.rs
```diff
@@ -95,10 +94,10 @@ pub(crate) fn get_project_file_walker_types(language: Option<SupportedLanguage>)
     let mut types_builder = TypesBuilder::new();
     types_builder.add_defaults();
     if let Some(language) = language {
-        types_builder.select(get_supported_language_name_for_ignore_select(language));
+        types_builder.select(language.name_for_ignore_select());
     } else {
-        for name_for_ignore_select in SUPPORTED_LANGUAGE_NAMES_FOR_IGNORE_SELECT.iter() {
```
helixbass, line 100: The top-level `SUPPORTED_LANGUAGE_LANGUAGES`/`SUPPORTED_LANGUAGE_NAMES_FOR_IGNORE_SELECT` can really be considered "private"/"implementation details" (and I did in fact make them not `pub`) since you can achieve eg here iterating over them by iterating over the main `ALL_SUPPORTED_LANGUAGES` and then accessing the corresponding relevant "field" via the helper methods (eg here `.name_for_ignore_select()`)

proc_macros/Cargo.toml
peterstuart: Ha, maybe it's a weird joke.

proc_macros/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+        let mut name: Option<Ident> = Default::default();
+        let mut variants: Option<Vec<Ident>> = Default::default();
+        while !input.is_empty() {
+            let key: Ident = input.parse()?;
+            input.parse::<Token![=>]>()?;
+            match &*key.to_string() {
+                "name" => {
+                    name = Some(input.parse::<Ident>()?);
+                }
+                "variants" => {
```
peterstuart, line 44: Whoa, interesting. I guess I haven't seen this kind of stuff in a proc-macro before.

proc_macros/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+}
+
+#[proc_macro]
+pub fn fixed_map(input: TokenStream) -> TokenStream {
+    let FixedMapArgs { name, variants } = parse_macro_input!(input as FixedMapArgs);
+
+    let collection_type_name = format_ident!("By{name}");
+    let all_variants_collection_name = format_ident!(
+        "ALL_{}",
+        name.to_string().to_plural().to_screaming_snake_case()
```
peterstuart, line 65: I love the name "screaming snake case", and it seems like most people haven't even heard of it 😄

proc_macros/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+                #values_struct_name::new(self)
+            }
+        }
+
+        pub struct #iter_struct_name<'collection, T> {
+            collection: &'collection #collection_type_name<T>,
+            next_index: usize,
+        }
+
+        impl<'collection, T> #iter_struct_name<'collection, T> {
```
peterstuart, line 149: I never really thing about how `Iterator` is implemented on structs that exist just for that purpose.

proc_macros/src/lib.rs
peterstuart: Ha yes, that's kind of ridiculous, but cool that it works.

proc_macros/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+impl Parse for ByFixedMapArgs {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let value_mapping_content;
+        bracketed!(value_mapping_content in input);
+        let mut value_mapping: HashMap<Ident, Expr> = Default::default();
+        while !value_mapping_content.is_empty() {
+            let key: Ident = value_mapping_content.parse()?;
+            value_mapping_content.parse::<Token![=>]>()?;
+            let value: Expr = value_mapping_content.parse()?;
+            value_mapping.insert(key, value);
```
peterstuart, line 309: Worth checking that it's not already present? (Unless that's handled elsewhere.)

src/language.rs
```diff
@@ -95,81 +32,79 @@ pub static ALL_SUPPORTED_LANGUAGES: BySupportedLanguage<SupportedLanguage> = {
...
+        CSharp => tree_sitter_c_sharp::language(),
+        Kotlin => tree_sitter_kotlin::language(),
+        Elisp => tree_sitter_elisp::language(),
+        Elm => tree_sitter_elm::language(),
+        Dockerfile => tree_sitter_dockerfile::language(),
+        Html => tree_sitter_html::language(),
+        TreeSitterQuery => tree_sitter_query::language(),
+        Json => tree_sitter_json::language(),
+        Css => tree_sitter_css::language(),
+        Lua => tree_sitter_lua::language(),
```
peterstuart, line 71: How would you feature-gate these (if you wanted to eventually), since you can't use the normal annotation in a macro. (Although I suppose you could make the macro understand that annotation?)

peterstuart (approved):

proc_macros/src/lib.rs
helixbass: Ya I was sort of "winging it" off of picturing the `macro! { key => value, other_key => other_value }` "invocation style"

proc_macros/src/lib.rs
helixbass: Right I haven't read that many implementations of them (eg of whatever the types are returned by `Vec::into_iter()` etc) but I think they just tend to be little "state machines"?

proc_macros/src/lib.rs
helixbass: Ah ya that makes sense, updated

src/language.rs
helixbass: > Although I suppose you could make the macro understand that annotation?

Ya per I think a comment on a subsequent PR (where I did `cpp` -> `c++`), if I parse into a baked-in `syn` type eg `syn::ExprArray` it parses annotations for me

But... this isn't real Rust syntax so couldn't use that directly

Maybe that would be an excuse to mimic how eg `syn` implements `.parse::<ExprArray>()` but have the "item" parsing be different or whatever

PR #54, opened 2023-07-12
Fixes from code review

helixbass:
In this PR:
- some fixes from code review from #50 

To test:
You should be able to pass `--language c++` as a command-line option
You should see languages printed in alphabetical order if you do eg `--help`

Based on `supported-languages-fixed-size-macro`


tests/languages.rs
```diff
@@ -52,7 +52,7 @@ fn test_objective_c_auto_language_ambiguous_query() {
         "objective_c_project",
         r#"
             $ tree-sitter-grep -q '(identifier) @c'
-            File "./example.h" has ambiguous file-type, could be ObjectiveC, C, or Cpp. Try passing the --language flag
+            File "./example.h" has ambiguous file-type, could be C, Cpp, or ObjectiveC. Try passing the --language flag
```
helixbass, line 55: Ah interesting that this still prints "Cpp", I guess this is using the `#[derive(Debug)]` implementation on `SupportedLanguage`, does that support declarative "customizing"/"overriding" attributes?

peterstuart: No, but strum does if you didn't mind using that: https://docs.rs/strum_macros/0.25.1/strum_macros/derive.Display.html (well technically `Display`, but that'd work I assume).

Or it seems pretty straightforward to add an implementation of `Debug` to your macro too?

peterstuart (approved):

tests/languages.rs
helixbass: I like the `strum` solution especially given that (a) we're already hard-coding "unusual" `#[derive(...)]`'s in the macro (`clap::ValueEnum`) and (b) it's already set up to "flow those per-variant attributes through" to the generated enum definition, updated

Gave a couple of the other languages slightly more expected formatted names too eg "C#" or here "Objective-C"

PR #57, opened 2023-07-13
Byte-offset option

helixbass:
In this PR:
- expose `-b`/`--byte-offset` option like `ripgrep`

To test:
If you pass `-b`/`--byte-offset` you should additionally see the byte offset (out of the entire file contents) at the beginning of output match lines

Based on `readme`


tests/output.rs
```diff
@@ -1113,3 +1120,67 @@ fn test_only_matching_multiline_overlapping_matches_starting_on_same_line() {
...
+    );
+}
+
+#[test]
+fn test_byte_offset_vimgrep() {
+    assert_sorted_output(
+        "rust_project",
+        r#"
+            $ tree-sitter-grep -q '(function_item) @function_item' -l rust --byte-offset --vimgrep
+            src/helpers.rs:1:1:0:pub fn helper() {}
```
helixbass, line 1168: This row:column:byte_offset is also what I see when running `ripgrep` with the combination of `--byte-offset` + `--vimgrep`

src/args.rs
```diff
@@ -99,6 +99,14 @@ pub struct Args {
...
     #[arg(short = 'o', long)]
     pub only_matching: bool,
+
+    /// Print the 0-based byte offset within the input
+    /// file before each line of output.
+    ///
+    /// If -o (--only-matching) is specified, print
+    /// the offset of the matching part itself.
+    #[arg(short = 'b', long)]
+    pub byte_offset: bool,
```
peterstuart, line 109: What's this used for?

tests/output.rs
```diff
@@ -1113,3 +1120,67 @@ fn test_only_matching_multiline_overlapping_matches_starting_on_same_line() {
...
+            $ tree-sitter-grep -q '(function_item) @function_item' -l rust --byte-offset
+            src/helpers.rs:1:0:pub fn helper() {}
+            src/stop.rs:1:0:fn stop_it() {}
+            src/lib.rs:3:14:pub fn add(left: usize, right: usize) -> usize {
+            src/lib.rs:4:63:    left + right
+            src/lib.rs:5:80:}
+            src/lib.rs:12:139:    fn it_works() {
+            src/lib.rs:13:159:        let result = add(2, 2);
+            src/lib.rs:14:191:        assert_eq!(result, 4);
+            src/lib.rs:15:222:    }
```
peterstuart, line 1138: So the second number here is the offset? I guess I've seen this before, but assumed without looking to hard it was the column. (Or is there another identically formatted grep output style that does work like that?)

peterstuart (approved):

tests/output.rs
helixbass: Ya it is kind of weird that it's the same "format" as eg `--vimgrep` mode where you have line number + column

So ya here the second # is the absolute byte offset

In the `test_byte_offset_vimgrep()` test below you can see that it is possible to combine those and get `row:column:byte_offset`

tests/output.rs
```diff
@@ -1113,3 +1120,67 @@ fn test_only_matching_multiline_overlapping_matches_starting_on_same_line() {
...
+
+#[test]
+fn test_byte_offset() {
+    assert_sorted_output(
+        "rust_project",
+        r#"
+            $ tree-sitter-grep -q '(function_item) @function_item' -l rust --byte-offset
+            src/helpers.rs:1:0:pub fn helper() {}
+            src/stop.rs:1:0:fn stop_it() {}
+            src/lib.rs:3:14:pub fn add(left: usize, right: usize) -> usize {
```
helixbass, line 1132: I'm seeing these tests fail in CI on Windows where it looks like the byte offset is bigger (by the row # - 1 bytes). That smells like "it's counting an additional `\r` for each line". But not totally clear on if/why the file contents (specifically the line endings) of the fixture files would be different on Windows?

So I think I'm going to rebase/re-target the other currently open PR against `master` and come back to this one

peterstuart: Is it possible that git does some line ending magic to get you native line endings on whatever OS you're on?

Maybe, depending on your config, it looks like: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings. Not sure what the default is.

helixbass: Nice ok I added `tests/fixtures/** eol=lf` to `.gitattributes` and it looks like it worked

That seems appropriate to me I guess I think that would just mean that someone editing something under `tests/fixtures/` on Windows wouldn't see "normal" Windows line endings?

helixbass: I guess that implies that we are no longer testing output when run against files with `\r\n` line endings though?

helixbass: Ok to avoid having to try and fully understand the implications of that, I just refined it to only impose that setting for the one fixtures directory for the `--byte-offset`-related tests (so all of the other existing tests will continue behaving the way they were before)

peterstuart: > That seems appropriate to me I guess I think that would just mean that someone editing something under tests/fixtures/ on Windows wouldn't see "normal" Windows line endings?

Yeah but I think modern editors can deal with that fine these days, so it probably doesn't make much difference?

PR #58, opened 2023-07-13
Fail on no captures

### LINKED ISSUE #44: Detect if there are no captures in query?
Not sure if there's a "provided" way to achieve this?

helixbass:
In this PR:
- fail "nicely" if the provided query contains no captures

Closes #44 

To test:
If you try passing a query with no captures in it eg `-q (function_item)` then it should print an error message (only once) that there must be at least one capture in the query and then exit with an error status code

Based on `byte-offset`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,18 +47,27 @@ struct CaptureIndex(OnceLock<Result<u32, ()>>);
 
 impl CaptureIndex {
     pub fn get_or_init(&self, query: &Query, capture_name: Option<&str>) -> u32 {
-        let mut did_mark_failed = false;
+        let mut failure_message: Option<String> = Default::default();
         self.0
             .get_or_init(|| match capture_name {
-                None => Ok(0),
+                None => match query.capture_names().len() {
+                    0 => {
+                        failure_message = Some(
+                            "query must include at least one capture (\"@whatever\")".to_owned(),
+                        );
+                        #[allow(clippy::unit_arg)]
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -47,18 +47,27 @@ struct CaptureIndex(OnceLock<Result<u32, ()>>);
...
+        let mut failure_message: Option<String> = Default::default();
         self.0
             .get_or_init(|| match capture_name {
-                None => Ok(0),
+                None => match query.capture_names().len() {
+                    0 => {
+                        failure_message = Some(
+                            "query must include at least one capture (\"@whatever\")".to_owned(),
+                        );
+                        #[allow(clippy::unit_arg)]
```
peterstuart, line 58: I looked up this link but I don't really understand what it's flagging here.

peterstuart (approved):

src/lib.rs
helixbass: Apparently it dislikes using something other than a literal `()` to say "`()`" but I didn't bother reading its explanation why it doesn't like that (sounded like it was something about a likely syntax mistake which is not the case here)

(and here I'm using `Default::default()` to mean `()`, because I am liking using `Default::default()` where it conceptually applies)

(stepping back a little bit, I'm not sure why I'm using `Result<_, ()>` here which smells like just an `Option<_>`, but that's how it already was and I saw that I was using `Default::default()` for the `()` elsewhere here so mimicked that)

PR #60, opened 2023-07-15
Proof-of-concept of exposing crate-level "push"-style API

helixbass:
In this PR:
- "proof-of-concept" of exposing a crate-level API that allows passing a callback that gets invoked for each matching tree-sitter AST node

Not in this PR:
- thinking about how to reasonably pass `Args`/configuration via a crate-level API
- tests
- docs for the crate-level API

To test:
Per the newly added example, you should be able to call the `tree_sitter::run_with_callback()` function (with a constructed `Args` instance) and do stuff in the provided callback that has access to the matched `tree_sitter::Node`, the file path where the match was found, and the source text for that file

Doing stuff via that crate-level API should be roughly as fast as via the command line (I assume)
(and performance via the command-line should still be about the same as what it was)

Based on `error-types`

### DIFF
--- a/examples/print_match_text.rs
+++ b/examples/print_match_text.rs
@@ -1,0 +1,13 @@
+use clap::Parser;
+use tree_sitter_grep::{run_with_callback, Args};
+
+fn main() {
+    let args = Args::parse_from(["tree_sitter_grep", "-q", "(function_item) @f"]);
+    run_with_callback(args, |node, file_contents, path| {
+        println!(
+            "Found match in {path:?}: {}",
+            std::str::from_utf8(&file_contents[node.byte_range()]).unwrap(),
+        );
+    })
+    .unwrap();
+}
--- a/src/args.rs
+++ b/src/args.rs
@@ -1,4 +1,5 @@
 use std::{
+    fs,
     path::{Path, PathBuf},
     sync::{Arc, Mutex},
 };
@@ -10,14 +11,17 @@ use termcolor::BufferWriter;
 
 use crate::{
     language::SupportedLanguage,
+    plugin::{get_loaded_filter, Filterer},
     printer::StandardBuilder,
     project_file_walker::{
         get_project_file_walker_types, into_parallel_iterator, WalkParallelIterator,
     },
     searcher::{Searcher, SearcherBuilder},
     use_printer::Printer,
-    NonFatalSearchError,
+    Error, NonFatalSearchError,
 };
+
+const ALL_NODES_QUERY: &str = "(_) @node";
 
 #[derive(Parser)]
 #[clap(group(
@@ -189,5 +193,24 @@ impl Args {
         non_fatal_errors: Arc<Mutex<Vec<NonFatalSearchError>>>,
     ) -> IterBridge<WalkParallelIterator> {
         into_parallel_iterator(self.get_project_file_walker(), non_fatal_errors)
+    }
+
+    pub(crate) fn get_loaded_filter(&self) -> Result<Option<Arc<Filterer>>, Error> {
+        Ok(get_loaded_filter(self.filter.as_deref(), self.filter_arg.as_deref())?.map(Arc::new))
+    }
+
+    pub(crate) fn get_loaded_query_text(&self) -> Result<String, Error> {
+        Ok(
+            match (self.path_to_query_file.as_ref(), self.query_text.as_ref()) {
+                (Some(path_to_query_file), None) => fs::read_to_string(path_to_query_file)
+                    .map_err(|source| Error::QueryFileReadError {
+                        source,
+                        path_to_query_file: path_to_query_file.clone(),
+                    })?,
+                (None, Some(query_text)) => query_text.clone(),
+                (None, None) => ALL_NODES_QUERY.to_owned(),
+                _ => unreachable!(),
+            },
+        )
     }
 }
--- a/src/bin/tree-sitter-grep.rs
+++ b/src/bin/tree-sitter-grep.rs
@@ -1,11 +1,11 @@
 use std::process;
 
 use clap::Parser;
-use tree_sitter_grep::{run, Args, RunStatus};
+use tree_sitter_grep::{run_print, Args, RunStatus};
 
 pub fn main() {
     let args = Args::parse();
-    match run(args) {
+    match run_print(args) {
         Ok(RunStatus {
             non_fatal_errors,
             matched,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,7 +13,7 @@ use ignore::DirEntry;
 use rayon::prelude::*;
 use termcolor::{BufferWriter, ColorChoice};
 use thiserror::Error;
-use tree_sitter::{Query, QueryError};
+use tree_sitter::{Node, Query, QueryError};
 
 mod args;
 mod language;
@@ -257,7 +257,7 @@ impl OutputContext {
     }
 }
 
-pub fn run(args: Args) -> Result<RunStatus, Error> {
+pub fn run_print(args: Args) -> Result<RunStatus, Error> {
     run_for_context(
         args,
         OutputContext::new(BufferWriter::stdout(ColorChoice::Never)),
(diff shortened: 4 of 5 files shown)


examples/print_match_text.rs
```diff
@@ -0,0 +1,13 @@
+use clap::Parser;
+use tree_sitter_grep::{run_with_callback, Args};
+
+fn main() {
+    let args = Args::parse_from(["tree_sitter_grep", "-q", "(function_item) @f"]);
```
helixbass, line 5: Per PR comment, seems non-ideal to force callers to "act as though they're providing a sequence of command-line arguments" in order to "configure" the behavior?

examples/print_match_text.rs
```diff
@@ -0,0 +1,13 @@
...
+    run_with_callback(args, |node, file_contents, path| {
+        println!(
+            "Found match in {path:?}: {}",
+            std::str::from_utf8(&file_contents[node.byte_range()]).unwrap(),
+        );
+    })
+    .unwrap();
```
helixbass, line 12: This should be getting back things like the non-fatal errors (or a fatal error if one occurred), might be interesting to "see that working"

src/args.rs
```diff
@@ -190,4 +194,23 @@ impl Args {
     ) -> IterBridge<WalkParallelIterator> {
         into_parallel_iterator(self.get_project_file_walker(), non_fatal_errors)
     }
+
+    pub(crate) fn get_loaded_filter(&self) -> Result<Option<Arc<Filterer>>, Error> {
```
helixbass, line 198: These were "smelling like `Args` methods" so moved them here

These both have some cost associated with them so I guess they shouldn't be thought of as "idempotent" or whatever - maybe the "loaded" in the method names gives that clue?

src/bin/tree-sitter-grep.rs
```diff
@@ -1,11 +1,11 @@
 use std::process;
 
 use clap::Parser;
-use tree_sitter_grep::{run, Args, RunStatus};
+use tree_sitter_grep::{run_print, Args, RunStatus};
 
 pub fn main() {
     let args = Args::parse();
-    match run(args) {
+    match run_print(args) {
```
helixbass, line 8: Naming?

src/lib.rs
```diff
@@ -250,31 +247,86 @@ impl<TSuccess> From<NonFatalSearchError> for Result<TSuccess, SingleFileSearchEr
...
-                path_to_query_file: path_to_query_file.clone(),
-            })?
-        }
-        (None, Some(query_text)) => query_text.clone(),
-        (None, None) => ALL_NODES_QUERY.to_owned(),
-        _ => unreachable!(),
-    };
-    let filter =
-        get_loaded_filter(args.filter.as_deref(), args.filter_arg.as_deref())?.map(Arc::new);
+pub struct OutputContext {
```
helixbass, line 250: Ok so `run_for_context()` is the shared entry point between the command-line vs crate-level API cases

The command-line case needs to "set up a `BufferWriter`" so I expressed that as a "context" object that the shared entry point is generic wrt

src/lib.rs
```diff
@@ -250,31 +247,86 @@ impl<TSuccess> From<NonFatalSearchError> for Result<TSuccess, SingleFileSearchEr
...
+    pub fn new(buffer_writer: BufferWriter) -> Self {
+        Self { buffer_writer }
+    }
+}
+
+pub fn run_print(args: Args) -> Result<RunStatus, Error> {
+    run_for_context(
+        args,
+        OutputContext::new(BufferWriter::stdout(ColorChoice::Never)),
+        |context: &OutputContext,
```
helixbass, line 264: So then the per-file "do the actual core searching stuff" callback gets access to the stuff it needs including that context object

src/lib.rs
```diff
@@ -250,31 +247,86 @@ impl<TSuccess> From<NonFatalSearchError> for Result<TSuccess, SingleFileSearchEr
...
+                .borrow_mut()
+                .search_path_callback::<_, io::Error>(
+                    query_context,
+                    path,
+                    |node: Node, file_contents: &[u8], path: &Path| {
+                        callback(node, file_contents, path);
+                        matched.store(true, Ordering::SeqCst);
+                    },
+                )
+                .unwrap();
```
helixbass, line 308: This mimics the `.search_path().unwrap()` above but it's presumably not good to just be `.unwrap()`'ing either one (this would represent eg an `io::Error` while doing the searching)

src/lib.rs
```diff
@@ -250,31 +247,86 @@ impl<TSuccess> From<NonFatalSearchError> for Result<TSuccess, SingleFileSearchEr
...
+         path: &Path,
+         query_context: QueryContext,
+         matched: &AtomicBool| {
+            get_searcher(args)
+                .borrow_mut()
+                .search_path_callback::<_, io::Error>(
+                    query_context,
+                    path,
+                    |node: Node, file_contents: &[u8], path: &Path| {
+                        callback(node, file_contents, path);
```
helixbass, line 304: A "push"-style API seemed like the easiest thing to expose at least to start with

src/searcher/mod.rs
```diff
@@ -212,6 +214,38 @@ impl Searcher {
         self.search_file_maybe_path(query_context, Some(path), &file, write_to)
     }
 
+    pub fn search_path_callback<P, TError: SinkError>(
```
helixbass, line 217: These sort of parallel the existing `Searcher` methods, just taking a callback instead of a `Sink` more or less

Didn't really think about trying to DRY up too much

src/searcher/mod.rs
```diff
@@ -297,6 +334,60 @@ impl Searcher {
...
+    }
+
+    fn run_with_callback(
+        &self,
+        query_context: QueryContext,
+        slice: &[u8],
+        callback: impl Fn(Node, &[u8], &Path),
+        path: &Path,
+    ) {
+        let mut query_cursor = QueryCursor::new();
```
helixbass, line 359: This also feels very non-DRY wrt `MultiLine::run()` or wherever this stuff happens in "command-line mode"

I've slightly thought about encapsulating this before and seemed tricky wrt eg lifetimes

src/searcher/mod.rs
```diff
@@ -297,6 +334,60 @@ impl Searcher {
...
+                );
+                match filter.as_ref() {
+                    None => Some(single_captured_node),
+                    Some(filter) => filter
+                        .call(&single_captured_node)
+                        .then_some(single_captured_node),
+                }
+            })
+            .for_each(|node| {
+                callback(node, slice, path);
```
helixbass, line 387: No sinks, no line-orientation, no printer, just give the callback basically the raw match data (ie the `tree_sitter::Node`) from the tree-sitter API

src/searcher/mod.rs
```diff
@@ -340,44 +431,36 @@ impl Searcher {
         self.config.passthru
     }
 
-    fn fill_multi_line_buffer_from_file<S: Sink>(&self, file: &File) -> Result<(), S::Error> {
+    fn fill_multi_line_buffer_from_file(&self, file: &File) -> io::Result<()> {
```
helixbass, line 434: This I just refactored to not be generic wrt a sink error type because it always returns `io::Result` and I thought I might be able to just use `io::Result` everywhere in the callback case (that ended up not being true so it also is generic wrt a `SinkError`)

(so basically things that call this just "more lazily" convert the `io::Error` -> `SinkError`)

examples/print_match_text.rs
```diff
@@ -0,0 +1,13 @@
+use clap::Parser;
+use tree_sitter_grep::{run_with_callback, Args};
+
+fn main() {
+    let args = Args::parse_from(["tree_sitter_grep", "-q", "(function_item) @f"]);
+    run_with_callback(args, |node, file_contents, path| {
```
helixbass, line 6: Naming?

src/bin/tree-sitter-grep.rs
peterstuart: (And that comment is related to naming because having to come up with various `run_something` names seems like the result of a weird split between the bin and lib to me.)

peterstuart (commented):
I didn't have time to do a very thorough review right now, but I left one comment/question inline about what the boundary between lib and bin should be, which seems important to figure out for the crate's interface. (But I don't have any great suggestions.)

src/bin/tree-sitter-grep.rs
helixbass: Ya agreed I think we're a little bit "weird-shaped"

I'm personally probably ok with the crate-level API feeling a bit like "hey yes this is hacked-onto something that is most naturally used via the command line"

That being said I think instantiating `Args` the way it's done here is by far _too_ hacked-on

PR #61, opened 2023-07-18
WIP: show heading + color

helixbass:
In this PR:
- mimic `ripgrep`'s heading/filename/line/column formatting (at least much more closely), eg showing a per-file "heading" and colors by default when outputting to a terminal

To test:
Now when running in a terminal by default you should see `ripgrep`-style output formatting (but with a different default set of colors)
If you eg redirect the output by doing `| cat -`, you should see different formatting that is more like our existing formatting but without showing line numbers (by default)
You should be able to use the newly exposed command-line arguments to modify those behaviors, eg use `--pretty` when redirecting the output to still get color/headings etc, use `--line-number` when redirecting the output to still get line numbers

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -3,14 +3,14 @@ use std::{
     sync::{Arc, Mutex},
 };
 
-use clap::{ArgGroup, Parser};
+use clap::{ArgGroup, Parser, ValueEnum};
 use ignore::{types::Types, WalkBuilder, WalkParallel};
 use rayon::iter::IterBridge;
-use termcolor::BufferWriter;
+use termcolor::{BufferWriter, ColorChoice};
 
 use crate::{
     language::SupportedLanguage,
-    printer::StandardBuilder,
+    printer::{default_color_specs, ColorSpecs, StandardBuilder, UserColorSpec},
     project_file_walker::{
         get_project_file_walker_types, into_parallel_iterator, WalkParallelIterator,
     },
@@ -18,6 +18,14 @@ use crate::{
     use_printer::Printer,
     NonFatalError,
 };
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, ValueEnum)]
+pub enum ColorChoiceArg {
+    Never,
+    Auto,
+    Always,
+    Ansi,
+}
 
 #[derive(Parser)]
 #[clap(group(
@@ -107,6 +115,73 @@ pub struct Args {
     /// the offset of the matching part itself.
     #[arg(short = 'b', long)]
     pub byte_offset: bool,
+
+    // This flag specifies color settings for use in the output.
+    //
+    // This flag may be provided multiple times. Settings are applied iteratively. Colors are
+    // limited to one of eight choices: red, blue, green, cyan, magenta, yellow, white and
+    // black. Styles are limited to nobold, bold, nointense, intense, nounderline
+    // or underline.
+    //
+    // The format of the flag is '{type}:{attribute}:{value}'. '{type}' should be
+    // one of path, line, column or match. '{attribute}' can be fg, bg or style.
+    // '{value}' is either a color (for fg and bg) or a text style. A special format,
+    // '{type}:none', will clear all color settings for '{type}'.
+    //
+    // For example, the following command will change the match color to magenta and
+    // the background color for line numbers to yellow:
+    //
+    //     tree-sitter-grep --colors 'match:fg:magenta' --colors 'line:bg:yellow' -q
+    // '(function_item) @f'
+    //
+    // Extended colors can be used for '{value}' when the terminal supports ANSI color
+    // sequences. These are specified as either 'x' (256-color) or 'x,x,x' (24-bit
+    // truecolor) where x is a number between 0 and 255 inclusive. x may be given as
+    // a normal decimal number or a hexadecimal number, which is prefixed by `0x`.
+    //
+    // For example, the following command will change the match background color to
+    // that represented by the rgb value (0,128,255):
+    //
+    //     tree-sitter-grep --colors 'match:bg:0,128,255'
+    //
+    // or, equivalently,
+    //
+    //     tree-sitter-grep --colors 'match:bg:0x0,0x80,0xFF'
+    //
+    // Note that the intense and nointense style flags will have no effect when
+    // used alongside these extended color codes.
+    #[arg(long)]
+    pub colors: Vec<UserColorSpec>,
+
+    #[arg(long/*, default_value_t = ColorChoiceArg::Auto*/)]
+    pub color: Option<ColorChoiceArg>,
+
+    // This is a convenience alias for '--color always --heading --line-number'.
+    //
+    // This flag is useful when you still want pretty output even if you're piping tree-sitter-grep
+    // to another program or file. For example: 'tree-sitter-grep -p -q "(function_item) @c" | less
+    // -R'.
+    #[arg(short = 'p', long)]
+    pub pretty: bool,
+
+    // This flag prints the file path above clusters of matches from each file instead of printing
+    // the file path as a prefix for each matched line.
+    //
+    // This is the default mode when printing to a terminal.
+    //
+    // This overrides the --no-heading flag.
+    #[arg(long)]
+    pub heading: bool,
+
+    // Don't group matches by each file.
+    //
+    // If --no-heading is provided in addition to the -H/--with-filename flag, then file paths will
+    // be printed as a prefix for every matched line. This is the default mode when not printing to
+    // a terminal.
+    //
+    // This overrides the --heading flag.
+    #[arg(long, hide = true, overrides_with = "heading")]
+    pub no_heading: bool,
 }
 
 impl Args {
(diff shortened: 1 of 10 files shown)


src/args.rs
```diff
@@ -190,4 +355,74 @@ impl Args {
...
+    fn file_separator(&self) -> Option<Vec<u8>> {
+        // if self.output_kind() != OutputKind::Standard {
+        //     return Ok(None);
+        // }
+
+        let (ctx_before, ctx_after) = self.contexts();
+        if self.heading() {
+            Some(b"".to_vec())
+        } else if ctx_before > 0 || ctx_after > 0 {
+            self.context_separator()
```
helixbass, line 405: This is what is responsible for the fact that now in some of the existing test diffs you'll see that there are `--` lines between files when passing a "context" option

(so we weren't matching `ripgrep`'s behavior there basically, now we are)

src/args.rs
```diff
@@ -190,4 +355,74 @@ impl Args {
...
+    }
+
+    fn with_filename(&self, paths: &[PathBuf]) -> bool {
+        if self.no_filename {
+            false
+        } else {
+            let path_stdin = Path::new("-");
+            self.with_filename
+                || self.vimgrep
+                || paths.len() > 1
```
helixbass, line 422: And this is causing it to not show the filename in the match output when you just pass a single file path (which is what `ripgrep` does)

src/printer/color.rs
```diff
@@ -4,16 +4,18 @@ use std::{error, fmt, str::FromStr};
 
 use termcolor::{Color, ColorSpec, ParseColorError};
 
-#[allow(dead_code)]
 pub fn default_color_specs() -> Vec<UserColorSpec> {
     vec![
         #[cfg(unix)]
-        "path:fg:magenta".parse().unwrap(),
+        "path:fg:green".parse().unwrap(),
```
helixbass, line 10: I forgot that I had `ripgrep`'s default colors overridden in a local config file

I can't with the default `ripgrep` colors they're disgusting

So this is "my" color scheme (don't remember where I got it from or if I made it up)

tests/languages.rs
```diff
@@ -8,7 +8,7 @@ fn test_swift() {
         "swift_project",
         r#"
             $ tree-sitter-grep -q '(value_argument) @c' --language swift
-            example.swift:2:    atPath: "native"
+            example.swift:    atPath: "native"
```
helixbass, line 11: Most of the existing test output changed because (a) those tests are running in "non-terminal-output" mode and (b) now in that mode we don't show line numbers by default (like `ripgrep`)

tests/output.rs
```diff
@@ -634,20 +769,22 @@ fn test_after_context() {
...
             $ tree-sitter-grep -q '(function_item) @f' -l rust --after-context 2
-            src/stop.rs:1:fn stop_it() {}
-            src/helpers.rs:1:pub fn helper() {}
-            src/lib.rs:3:pub fn add(left: usize, right: usize) -> usize {
-            src/lib.rs:4:    left + right
-            src/lib.rs:5:}
-            src/lib.rs-6-
-            src/lib.rs-7-#[cfg(test)]
+            src/stop.rs:fn stop_it() {}
+            --
```
helixbass, line 773: Per above here are some new `--` separators between files when passing a "context" option

tests/shared.rs
```diff
@@ -193,6 +194,9 @@ pub fn assert_non_match_output(fixture_dir_name: &str, command_and_output: &str)
         .success()
         .stdout(predicate::function(|stdout: &str| {
             let stdout = massage_error_output(stdout);
+            if stdout != output {
+                print_diff(&stdout, &output, " ");
```
helixbass, line 198: This was a hacky approximation of a "pretty assertion diff" that I'm kind of inclined to keep in (it only kicks in if the test is failing)?

tests/output.rs
```diff
@@ -1190,8 +1343,358 @@ fn test_byte_offset_only_matching() {
...
+    );
+}
+
+#[test]
+fn test_pretty() {
+    assert_sorted_output(
+        "rust_project",
+        r#"
+             $ tree-sitter-grep -q '(function_item) @c' -l rust --pretty
+             [0m[1m[32msrc/stop.rs[0m
```
helixbass, line 1641: You could picture having a "cuter" test-syntax for this

But not sure it's worth it?

Surprisingly this passed on Windows too I guess they use the same terminal color escape codes or whatever?

tests/shared.rs
```diff
@@ -141,11 +142,11 @@ fn massage_windows_line(line: &str) -> String {
 }
 
 fn strip_trailing_carriage_return(line: &str) -> Cow<'_, str> {
-    regex!(r#"\r$"#).replace(line, "")
+    regex!(r#"\r((?:\u{1b}\[\d+m)*)$"#).replace(line, "$1")
```
helixbass, line 145: I was seeing terminal color escape codes _after_ `\r`'s (not sure if that's expected?) so this handles that

tests/shared.rs
```diff
@@ -141,11 +142,11 @@ fn massage_windows_line(line: &str) -> String {
...
 }
 
 fn normalize_match_path(line: &str) -> Cow<'_, str> {
-    regex!(r#"^[^:]+[:-]\d+[:-]"#)
+    regex!(r#"^(?:\u{1b}\[\d+m)*[a-zA-Z_\-\\/]+\.(?:rs|js|tsx)"#)
```
helixbass, line 149: And here this regex for finding "paths" (for changing `\` -> `/` on Windows) keeps getting re-worked, now it really can't really on "context" at all (since eg heading lines are just paths by themselves)

So just expressed it in a way that works for all current tests with filename output

.github/workflows/main.yml
```diff
@@ -24,5 +24,10 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
       - run: cargo clippy --all-targets --tests -- -D warnings
       - run: cargo test
+        if: matrix.os == 'ubuntu-latest'
+        env:
+          TERM: linux
```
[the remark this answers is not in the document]
peterstuart: That seems reasonable, but maybe that suggests that you should be able to control environment variables in tests?

src/args.rs
```diff
@@ -107,10 +119,137 @@ pub struct Args {
...
+
+    /// Display the file path for matches.
+    ///
+    /// This is the default when more than one file is searched. If --heading is
+    /// enabled (the default when printing to a terminal), the file path will be
+    /// shown above clusters of matches from each file; otherwise, the file name
+    /// will be shown as a prefix for each matched line.
+    ///
+    /// This flag overrides --no-filename.
+    #[arg(short = 'H', long)]
```
peterstuart, line 213: Assuming you took this from ripgrep, but what's the "H" mean? "heading-ish"?

src/args.rs
```diff
@@ -107,10 +119,137 @@ pub struct Args {
...
+    #[arg(long)]
+    pub column: bool,
+
+    #[arg(long, hide = true, overrides_with = "column")]
+    pub no_column: bool,
 }
 
 impl Args {
-    fn use_paths(&self) -> Vec<PathBuf> {
+    pub fn use_paths(&self) -> Vec<PathBuf> {
```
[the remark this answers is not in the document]
peterstuart: You could probably write a macro which does the once-cell thing that you want to the function it annotates?

src/printer/color.rs
peterstuart: I find the highlights pretty gross when running in my terminal using my color scheme (I assume this is somehow using the colors I have specified there?):

Screenshot 2023-07-20 at 5 12 37 PM

But it's readable 🤷🏻

tests/output.rs
peterstuart: Huh, I find that pretty surprising given how incompatible the terminals are in most other ways. (But they do somehow run git and whatever else, so maybe they are trying to be compatible.)

peterstuart: > You could picture having a "cuter" test-syntax for this

I can't really 😄 — what were you thinking?

> But not sure it's worth it?

Yeah, doesn't seem like it's worth a ton of effort.

tests/shared.rs
peterstuart: That's not where I'd have guessed they'd go, but why not 🤷🏻

tests/shared.rs
peterstuart: What happens when color command codes are in the diff?

peterstuart (approved):
Nice, looks good 👍🏻

src/args.rs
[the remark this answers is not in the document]
peterstuart: Ah yeah, I didn't really think about how it'd have to be per-instance.

That seems like a reasonable approach to me (the once cell field). I wonder if clap exposes some option to call a constructor for `Args` so that you could at startup calculate this stuff rather than having to do it lazily. (It's not like it's not going to be used, so having to do work to make it smartly lazy seems kind of silly.)

src/printer/color.rs
peterstuart: Ha yes, 100% of people want it to be configurable, I guess, and I definitely couldn't be bothered to pass `--colors` every time I used it.

